PR #4, opened 2017-11-12
Update build

davidyuk:


muxe:
Thanks for including a build process. Can you update the readme on the usage because I can't get it to work.
The entry point in the package.json points to "dist/aepp-components.esm.js" which is in .gitignore and therefore is not included in the npm module.
How and where would you use the "aepp-components.esm.js" and the "aepp-components.js".
Thanks

davidyuk:
I decided to remove bundle for browser for simplifying. So, there are only `aepp-components.js`, it exports installation function for all packages as before, `aeHelperMixin`, and all components separately as single-file components.
Also, I have added `prepublish` script for using with `npm publish`, and published this package https://www.npmjs.com/package/aepp-components-davidyuk.
`aepp-components.js` will be included in an NPM packages, since I have added `dist` to array `files` of `packages.json`.
Readme is updated with an example of using single-file component.

PR #8, opened 2017-11-28
Add AeHr, AeHrButton, AeHrProgressbar components

davidyuk:
@dadaphl why you have removed AeHr and AeHrButton in 914314d57252fb4d2e3e07c7734b70c5e22a7be3 as unfinished? These components are already used by the Wall app.

### DIFF
--- a/src/components/aeHr/aeHr.md
+++ b/src/components/aeHr/aeHr.md
@@ -1,0 +1,7 @@
+```js
+<ae-hr />
+```
+
+```js
+<ae-hr branded />
+```
--- a/src/components/aeHr/aeHr.vue
+++ b/src/components/aeHr/aeHr.vue
@@ -1,0 +1,34 @@
+<template>
+  <hr class="ae-hr" :class="{ branded }" />
+</template>
+
+<script>
+  /**
+   * Displays an horizontal rule
+   */
+  export default {
+    name: 'ae-hr',
+    props: {
+      /**
+       * Use brand color instead of default
+       */
+      branded: { type: Boolean, default: false }
+    }
+  }
+</script>
+
+<style lang="scss" scoped>
+  @import "../variables";
+
+  .ae-hr {
+    background-color: $silver;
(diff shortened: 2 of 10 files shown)


dadaphl:
@davidyuk Hello,

sorry, It was a quick decision because of a lack of time to refactor them.

The design team and I were thinking a lot about structural requirements for components in the past days. We figured that composability should be a key requirement for our components. In general we should focus on components that are simple, yet work in as many applications as possible and well together with other components. Unfortunately we are still in an early stage of developing this design system so there is no clear guideline yet.

I would like to avoid, that the components repository becomes cluttered with components that only work for one specific use case (or one of our Aepps). Naming is also important. Components should be named after UI elements not concepts of a Aepp (for example AeCategory is actually a AeLabel).
I think also for other developers to use the components there needs to be some kind of consistency in CSS classnames and so on. 

Despite the fact that the components from you PR are well built, they don't seem fit in the overall design system. We can discuss this in detail and develop the guidelines together.

We are going to update the readme as soon as possible.

## AeHr:

I think there is no real benefit for a AeHr component in the repository, it doesn't really do anything 'better' than a `<hr>` element and offers little opportunity for composability at the moment.

Maybe we can refactor it as a AeSeparator or AeSpacer (name can be different) that can be used for different kinds of spacings and separator styles. We could also refactor the AeFilterSeparator into the same component. 

## AeHrButton:

We decided, that there will be only be a general AeButton component in the repository, a AeHrButton (or AeHeaderButton) component is to specific. It would be better if one composes the desired UI elements from more general components. We could give a AeSeparator a slot in which one can place a AeButton (or something else)

```
<ae-separator type='horizontal'>
  <ae-button size='small'>example</ae-button>
</ae-separator>
```

## AeHrProgressbar

We should call it AeProgress and there could be a type='bar' prop? Maybe could also benefit from refactoring parts in a mixin and resuse it for a AeLoader (spinner) component.


Cheers

PR #9, opened 2017-11-28
Fix AeHeader

davidyuk:
Broken in c56571ae39d193b85e0b17dafab4d2c686024229


dadaphl:
I don't think it's a good idea to include the router in a component. We should make the components independent from other Vue plugins like router or store. Please consider rewriting this so you can achieve the desired behaviour in your application with slots, props or events.

PR #12, opened 2017-12-06
new design like in sketch

dadaphl:
screen shot 2017-12-06 at 18 34 46


dadaphl:
I have used this technique since a long time and forgot why, maybe it became obsolete. You could also call it [Cargo Cult Programming](https://en.wikipedia.org/wiki/Cargo_cult_programming)

PR #13, opened 2017-12-06
aeButton type:plain has no box-shadow

dadaphl:
```
    <ae-button @click='close' size='small' type='plain'>
      <ae-icon slot='icon' name='close'/>
    </ae-button>
```
## before
screen shot 2017-12-06 at 18 48 27

## after
screen shot 2017-12-06 at 18 48 52


dadaphl:
true

PR #14, opened 2017-12-07
update `new components submission guidelines` first draft

dadaphl:
Please discuss and expand/correct this guidelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Contributions to the collection are welcome! If you wrote a cool component that 
 
 ### Component Purpose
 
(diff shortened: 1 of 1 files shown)


dadaphl:
you can find a rendered version here

https://hackmd.io/GwZghgRgnAJqC0YDGBTAZvALCsGoAYB2FeAVgCYwLyCRyY0g?view

acdbaykal:
I'm not so sure about

> Add all modifier classes (that represent type, state or similar things) of the component to the main wrapping element.

I think you should use modifier with each element effected by the state.So 

`.ae-progress-indicator._size_large {
	.progress-icon {...}
	.progress-label {...}
}`

would be more like

`.ae-progress-indicator._size_large {}
 .progress-icon._size_large {}
 .progress-label._size_large{}
`
This reduces the specificity of your selectors and makes sub elements less dependent on their parent container. If you wanted pull out some elements into a new component and had used the method you suggested, you'd probably have to rewrite most of the css, because the parent element is not part of the new component. But if you let each child element handle it's state it becomes easier to reuse the css.

The second issue I have is not mentioning testing. We are shamefully guilty about that ourselves. But every component should be submitted with working tests.

PR #15, opened 2017-12-12
"Sandbox" Vue app for local development :) 

evvvritt:
`npm run dev`

- sets up a local webpack server from the /sandbox directory
- /sandbox router and components are ignored

**to do**
- figure out proper webpack setup to have index.html file inside /sandbox


davidyuk:
Can't we use `vue-styleguidist` server for this purpose?

evvvritt:
@davidyuk Ah maybe! Yeah, wasn't familiar with vue-styleguide. Though maybe a benefit of the sandbox is being able to play with different wrappers for components?

davidyuk:
You can try any combination of components in `<componentName>.md` files, like [here](https://github.com/aeternity/aepp-components/blob/master/src/components/aeFilterList/aeFilterList.md). In local development I also connect my version of `aepp-components` to an app project by `(npm link)[https://docs.npmjs.com/cli/link]`.

davidyuk:
Usage of `vue-styleguidist` has a downside in that some components look better when they wrapped with AeMain components, but styleguidist doesn't do it.

evvvritt:
Ok, let's go with vue-styleguide then. If I have time I will setup and submit a new PR.

PR #16, opened 2017-12-14
aeModal – update design, "closeable" prop, scrollable on overflow 

evvvritt:


evvvritt:
New UI – https://cl.ly/1X2j2J0Y1d3v

davidyuk:
I think `closeable` should be renamed to `showCloseButton` and it shouldn't affect on emitting of `close` event. Because this event is also emitted on click on the overlay and it can be used even in your case: it can do the same that "cancel" button does.
Also, the transformation of a modal window to full-screen mode on mobile is very useful in Wall, Response apps. Should be possible to enable it by property of the AeModal component.

evvvritt:
Is there a layout I can use to reference the full-screen mode? Agree about `showCloseButton` and will implement by Tuesday of next week.

davidyuk:
Dialog with full-screen mode on mobile is used in [response.aepps.com](https://response.aepps.com/) (should click the add button on the top right side). The layout of question creation dialog is [here](https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5a1d6c4e72a2df33370f0f7d).

evvvritt:
@davidyuk Implemented Denis's feedback in my latest commit. Please review again for merge. Thanks!
(cc @emil-apeunit)

dadaphl:
@davidyuk, @okwme  Is there a reason, why this has neither been merged nor closed, nor improved?

dadaphl:
I guess this PR is not important, so i'm closing it.

PR #17, opened 2017-12-14
fix jest test error

okwme:


muxe:
I assigned denis as well because it's his original code.
also the build still fails because of linter errors

davidyuk:
Also possible `require.context` is replaceble by [require-all](https://www.npmjs.com/package/require-all) package.

PR #22, opened 2018-01-08
Add components

davidyuk:


dadaphl:
I guess you want to rename label to badge to not confuse with the <label> element (like bootstrap changed). good idea! What i don't understand is, why is there still a `ae-label` component? Shouldn't it be deleted then?

davidyuk:
New `AeLabel` component is an implementation of `<label>`, this is used in Wall and Response aepps.
screenshot from 2018-01-11 16-04-52
`AeLoader` is used in Response aepp and can be used in Vote aepp.
screenshot from 2018-01-11 16-06-15

PR #23, opened 2018-01-17
Add AeModalLight component

davidyuk:
screenshot from 2018-01-17 17-55-05
https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5a27ffe4f9a38cac5e8ed165

I decided to implement it as a separate component because otherwise, code of AeModal becomes too complicated. Maybe will be better to remove fullscreen-on-mobile functionality of AeModal, but this functionality has a quite successfully usage in the Response app.

#16

### DIFF
--- a/src/components/aeButton/aeButton.js
+++ b/src/components/aeButton/aeButton.js
@@ -41,6 +41,10 @@ export default {
     invert: {
       type: Boolean,
       default: false
+    },
+    uppercase: {
+      type: Boolean,
+      default: false
     }
   },
   computed: {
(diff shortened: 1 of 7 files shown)


dadaphl:
I think it would be better if it had a distinguishable name. There is a hint In the styleguide that this things could be called `alerts/confirms/prompts`

suggestion:

`aeAlert`
`aePrompt`
`aeConfirm`

davidyuk:
`alert`, `confirm`, `prompt` has a well-defined meaning in JS, any of them in name of this component can be confusing, as far as all of them can be implemented using this component. Probably will be better to rename `AeModal` to `AeModalPage` and then this component rename to `AeModal`.

PR #24, opened 2018-01-19
AeButton: added block property

Philip-Nunoo:
Added property to make buttons block.

```
<div>
  <ae-button block>Block buttons that span full width</ae-button>
</div>
```


dadaphl:
do we really need this prop? what about, if we let the parent component control the display and width of a child component?

davidyuk:
I need a property that allows aligning label inside button. If styles of parent component are scoped it is impossible to do it without defining an unscoped block.
https://github.com/aeternity/aepp-identity/blob/6e7d406a62ce4b23d0694f88c97437b1bd88eff7/src/pages/Intro.vue#L79-L83

PR #25, opened 2018-01-19
Ae text input

acdbaykal:
These are three input field based components with unit tests and style guide examples:
- aeTextInput is the simplest. It provides an input and a clear button. Clicking the button however only sends out an clearRequest event together with the current value of the input. the embedding environment has to do the actual clearing by setting the value property to ''.
- aeValidatedTextInput enhances the input with an interface for validation. It differentiates between validation during typing and validation after the input has lost focus. The functions provided through the validateOnBlur and validateOnInput are run and their result (return value) is passed through a validation event
- aeTextInputMolecule combines validations with displayed validation result messages. The errorMessages property should contain validation error messages and a key for the message. Like {
     "MORE_CHARS": "At least 8 characters are required",
    "INVALID_CHARS": "Your text contains invalid chracters"
}

validateOnBlur and validateOnInput properties than should  return the error codes in errorMessages, if the current value is invalid.


dadaphl:
Maybe we should discuss, if we want to introduce 'Atomic Design Methodology'. I think it's not a bad idea, but then we should make it a mandatory for all components. The naming of the components should stay consistent. Also the guide needs to be updated. Its not obvious to everyone what a 'molecule' is.

acdbaykal:
Well be the [atomioc design definition](http://bradfrost.com/blog/post/atomic-web-design/) all of these components ate already molecules. Just didn't want to name aeTextInputMolecule something like aeTextInputWithValidationErrorMessages.

src/components/aeTextInputMolecule/aeTextInputMolecule.vue
```diff
@@ -0,0 +1,39 @@
...
+      <span
+        class="top-message _type_valid"
+        ref="validMessage"
+        v-if="!!_validMessage"
+      >
+        {{_validMessage}}
+      </span>
+    </div>
+    <ae-validated-text-input
+      :inputId="internalInputId"
```
davidyuk: `_uid` can be used instead of `internalInputId`, probably not very good to use internal vue api, but it simpler.

package.json
```diff
@@ -23,7 +24,8 @@
     "ethjs-unit": "^0.1.6",
     "normalize.css": "^7.0.0",
     "numeral": "^2.0.6",
-    "string-loader": "0.0.1"
+    "string-loader": "0.0.1",
+    "unit": "0.0.1-alpha"
```
davidyuk: I can't figure out how this package is used.

src/components/aeTextInputMolecule/aeTextInputMolecule.vue
acdbaykal: Didn't know about that. I think it's actually better. Because Vue will ensure that the ids are unique. Right know it's just probable that they are unique.

package.json
acdbaykal: fixed

src/components/aeTextInputMolecule/aeTextInputMolecule.vue
acdbaykal: fixed

davidyuk:
I have no experience with Atomic Design Methodology. But the base idea to construct components from existing ones seems reasonable. For now, I propose to avoid `Molecule` by renaming `AeTextInputMolecule` to `AeTextInputLabeled` or so.

We don't need to implement our own validation for form fields, I propose to use `vee-validate` package, or just make our components compatible with this package for using in final aepps.

Perhaps aepp developer should manually connect AeFormField component (not implemented yet) with AeLabel by `vee-validate`. This package has simple requirements for input-components, we should just not to break them.

acdbaykal:
@davidyuk,

I'm not happy with what {{vee-validate}} provides, for the following reasons:

- js code is the better  way to express validation rules. With the vee-validate way you need to learn all these rule names and what arguments they take. It's overly complicated and not event as flexible as what I implemented.

- vee-validate assumes you will be using the vue-i18n plugin.

- vee-validate does not allow you to perform different validations on input and blur events. Usually one wants to apply softer validation rules while the user is still typing. I.e. while typing in an a e-mail address  the value should only be invalid if an invalid character was typed. However when blured the whole formatting of an email should be considered.

davidyuk:
- It is possible to define [custom validation rules](http://vee-validate.logaretm.com/rules.html#custom-rules) in js code and attach them in a declarative way to input fields. Yes, `vee-validate` already provides some validation rules that can be used [like here](https://github.com/aeternity/aepp-identity/blob/bfb8b23ed3c678ba9768eb1296b942e79aae917e/src/pages/Login.vue#L14). I think to use it easier then define own validation.
- Usage of `vue-i18n` is optional according to http://vee-validate.logaretm.com/#configuration (`i18n` option).
- Seems that it is really impossible to perform different validations on input and blur events, but it is possible to debounce validation by [`delay`](http://vee-validate.logaretm.com/#configuration) global option or [`data-vv-delay`](http://vee-validate.logaretm.com/api.html#data-attributes) attribute or disable validation on input event by [`events`](http://vee-validate.logaretm.com/#configuration) global option or [`data-vv-validate-on`](http://vee-validate.logaretm.com/api.html#data-attributes) attribute, also it is possible to run validation only on blur event by [`v-model.lazy`](https://github.com/baianat/vee-validate/issues/338).

acdbaykal:
@davidyuk ,

- Pure, mostly single line  functions are plain, simple and flexible.  And indeed typing the rules in `vee-validate` is easy, I agree. But then there all these frameworkish  things `vee-validate` does like injecting a `v-validate` property into input elements. I don't want to attach `vee-validate` as a dependency to the components and impose it onto others.

- As a principle validation has no business with localisation. `vee-validate` should not provide any defauIt messages or favor one i18n solution over others.  It should just provide an error code to the outside and the rest of the code just does what it does with it.

- What you describe there is inadequate. I work on another project which is centered around forms. We do a lot of validation in that project. Therefore I can tell you, there are many use cases where a separation of blur and input validation is desirable. And all those attributes you listed, make me dislike `vee-validate` even more. It is simply bloated and cumbersome. 

@dadaphl, I doubt we'll come to an agreement on this issue. Would you please be the referee and make the decision.

davidyuk:
I don't propose to add `vee-validate` as a dependency to the component repository. I think we shouldn't build our own validation. Instead of `vee-validate`, an aepp developer can choose [`vuelidate`](https://www.npmjs.com/package/vuelidate) (this package doesn't provide any default messages) or write validation code by himself.

acdbaykal:
@davidyuk,
the point of this PR is to implement an input with validation messages. I don't think this is possible without implementing some kind of validation logic. However I suggest to refactor `LabelledTextInput` and make the input a `<slot/>`. In order to show validation messages the input than has to communicate its validation result through a validation event (just like `aeValidatedTextInput` does). But one can use any input which abides by this contract. `aeValidatedTextInput` is then an implementation you can choose to use, but are not required to. Would this be a sufficient solution for you?

acdbaykal:
Further refactoring:
- drop aeTextInput
- make the input in aeValidatedTextInput a `slot`. Any component with a value prop and blur and input events can be used here.
- rename aeValidatedTextInput into aeInputValidation

acdbaykal:
@dadaphl,

I need a decision on this one. How do we pursue? Shall I close the PR, shall I use a library like `vuelidate` for validation or shall I implement my suggested refactorings and merge.

dadaphl:
The whole conversation is a bit hard to follow, I checked out the code of the HEAD of this branch and decided its ok. In general i think we should have the components as light as possible and not bundle a external validation solution. it should be easy to hook a 3rd party non-opinionated validation library.

davidyuk:
I think this PR is merged too early, it duplicates the functionality of AeLabel and AeInput.

acdbaykal:
for once I agree ;). When the suggested refactorings were implemented it, one could use AeInput with the other components and the redundancy would be resolved.

For clarity sake I will summarize my suggestions here:
- aeTextInput will be removed
- aeInputValidation wraps any component with a value property and the focus, blur and input events. It sends out a validation event each time a validation has occurred. If an developer wants to use a different validation strategy/framework he/she may write a component which also sends out validation events and replace aeInputValidation
- aeLabelledInput bundles AeLabel  with the validation event. It takes an errorMessages prop, which allows enables the  localization of validation errors.
- aeInputBundle: provides a default configuration using aeInput.

@dadaphl , @davidyuk, are you ok with this?

acdbaykal:
@dadaphl , @davidyuk,
I was disheartened when I learned that it is not possible to handle events of slot elements. However I kept trying and researching and got to a promising state at commit  ad126c9fee9b4814b6f7089474ad42507da08c08. 

`
new Vue({
  data:{
    inputValue: 'Initial value',
    errMsg:''
  },
 template: `
  <div>
      <ae-validated-text-input
        @validation="onValidationResult"
        :validateOnBlur="validate"
        :validateOnInput="validate"
      >
        <ae-input :value="inputValue" placeholder="Test input"/>
      </ae-validated-text-input>
      <p>{{errMsg}}</p>
  </div>
  `,
  methods: {
      validate(value){
        const inValid = /\D/.test(value)
        if(inValid){
         return 'Only digits are allowed'
        }
      },
      onValidationResult(result){
        this.errMsg = typeof result === 'string' ? result : ''
      }
  }
})
`

works. If you use input instead of ae-input in the example there are currently still some bugs

src/components/aeAddressInput/aeAddressInput.scss
```diff
@@ -1,6 +1,15 @@
 @import '../variables.scss';
-label {
+@import '../_extendables.scss';
+
+input{
+  @extend %input;
```
davidyuk: `aeAddressInput.scss` is imported to `aeAddressInput.vue` without `scoped` atribute and this changes behavour of all inputs in project.

PR #27, opened 2018-01-20
Ae banner

Philip-Nunoo:
Add banner component


src/components/aeBanner/aeBanner.js
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @public
+     */
+    hideBanner () {
+      if (!this.show) {
+        return
+      }
+      this._clearTimer()
+      this.show = false
+    },
```
davidyuk: Is this component visible or not should be controlled outside of this component. This component should be a dumb as AeNotification.
[Here](https://github.com/aeternity/aepp-response/blob/5ff8b09d4f48c2126f4fef8085911df93ccdc4cf/src/store/modules/response.js#L251-L256) is an example of controlling such component from Vuex store.

src/components/aeBanner/aeBanner.js
```diff
@@ -0,0 +1,67 @@
...
+  name: 'ae-banner',
+  data: function () {
+    return {
+      show: false
+    }
+  },
+  props: {
+    'options': {
+      type: Object
+    }
```
davidyuk: This can be just `options: Object`.

src/components/aeBanner/aeBanner.js
```diff
@@ -0,0 +1,67 @@
...
+        clearTimeout(this.timers)
+      }
+    }
+  },
+  computed: {
+    hasRightItem () {
+      return !!this.$slots.right
+    },
+    allOptions () {
+      return Object.assign({}, defaultOptions, this.options)
```
davidyuk: Will be more clear to define two properties: `timeout` as Number and `auto-destroy` as Boolean with proper default values.

davidyuk: Or one property `auto-destroy-timeout` and if it is not set then don't hide a banner.

src/components/aeBanner/aeBanner.scss
```diff
@@ -0,0 +1,46 @@
...
+  right: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 19000;
+
+  @include tablet {
+    width: 748px;
+  }
+
+  @include desktop {
+    width: 880px;
```
davidyuk: This constant is available in `variables.scss`.

src/components/aeBanner/aeBanner.scss
```diff
@@ -0,0 +1,46 @@
...
+  position: fixed;
+  top: 40px;
+  left: 50%;
+  right: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 19000;
+
+  @include tablet {
+    width: 748px;
+  }
```
davidyuk: As I see from Zeplin AeBanner don't have special behavior in tablet layout, it just shouldn't be wider than 880px in all layouts.

src/components/aeBanner/aeBanner.scss
```diff
@@ -0,0 +1,46 @@
...
+$desktop-width: 1024px;
+
+@mixin tablet {
+  @media (min-width: #{$tablet-width}) and (max-width: #{$desktop-width - 1px}) {
+    @content;
+  }
+}
+
+@mixin desktop {
+  @media (min-width: #{$desktop-width}) {
```
davidyuk: `$desktop-width` shouldn't be wrapped with `#{...}` here.

src/components/aeBanner/aeBanner.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <div v-if="show" class="ae-banner">
+    <div class="inner">
+      <slot />
+      <div v-if='hasRightItem' class="right">
```
davidyuk: Is it important to hide this node is `right` slot is not presented?

src/components/aeBanner/aeBanner.js
Philip-Nunoo: Having two separate properties would be well enough. As one may only wish to change the timeout.

src/components/aeBanner/aeBanner.js
Philip-Nunoo: Yes is controlled via the **showBanner** and **hideBanner** methods

Philip-Nunoo:
We could remove aeNotification later if that is not necessary anymore.

src/styles/_mixins.scss
```diff
@@ -0,0 +1,13 @@
+@import "./../components/variables";
```
davidyuk: We don't need to use underscores in `.scss` file names, because we don't have an auto import of them.

src/components/aeBanner/aeBanner.js
davidyuk: I mean, this component should be stateless.

src/components/aeBanner/aeBanner.vue
```diff
@@ -2,8 +2,8 @@
   <div v-if="show" class="ae-banner">
     <div class="inner">
       <slot />
-      <div v-if='hasRightItem' class="right">
-        <slot name="right" />
+      <div class="right">
+        <slot name="right"  class="right" />
```
davidyuk: `class="right"` seems useless here because you have already wrapped it in this class.

dadaphl:
i think, we should get rid of AeNotification, it violates the "Name it after what users will see, not what it represent in your application." rule. So AeBanner is a good idea because its a visual component. 

Please check if there is functionality from AeNotification that needs to be ported to AeBanner

Like @davidyuk wrote, component visibility should be controlled outside of this component. This component should be a dumb as AeNotification.

Philip-Nunoo:
The component (AeBanner) is currently controlled outside with the methods ```showBanner`` and ```hideBanner```. The state value being passed only determines the initial state when initialized.

src/components/aeBanner/aeBanner.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <div class="ae-banner">
+    <div class="inner">
+      <slot />
+      <slot name="right">
+        <ae-button
```
dadaphl: That should be a optional button. also i think the parent component should provide it via slot. there are many designs without a `cancel` button.

src/components/aeNotification/aeNotification.js
```diff
@@ -1,48 +0,0 @@
-import AeButton from '../aeButton/aeButton.vue'
```
dadaphl: I'm confused, I thought we wanted to merge aeNotification with aeBanner. This branch just deletes aeNotification functionality. Please get in touch with Ray how to merge this two components in compliance with the current design.

src/components/aeBanner/aeBanner.scss
```diff
@@ -0,0 +1,22 @@
+@import "../variables";
+@import "./../../styles/mixins";
+
+.ae-banner {
+  width: 355px;
```
dadaphl: i think the banner should take up all horizontal space

src/components/aeBanner/aeBanner.vue
Philip-Nunoo: ok this was in tune with the old notification component.

src/components/aeNotification/aeNotification.js
Philip-Nunoo: The idea was to replace aeNotification with aeBanner, keeping most of the API and changing the design.

Philip-Nunoo:
@dadaphl @davidyuk Since this has been recreated here #42 I am closing this PR. And possibly request for the branch to be deleted as well.

PR #34, opened 2018-01-26
add aeDivider component

Philip-Nunoo:
This component is used in the new ae-identity card.

### DIFF
--- a/src/components/aeDivider/aeDivider.js
+++ b/src/components/aeDivider/aeDivider.js
@@ -1,0 +1,27 @@
+const aeDividerTypes = [
+  'boring',
+  'normal',
+  'exciting',
+  'dramatic'
+]
+
+export default {
+  name: 'ae-divider',
+  props: {
+    type: {
+      type: String,
+      validator: (value) => aeDividerTypes.find(e => e === value),
+      default: 'normal'
+    }
+  },
+  computed: {
+    typeModifier () {
+      return `ae-divider ${this.type}`
+    },
+    cssClass () {
+      return [
+        this.typeModifier
+      ]
+    }
+  }
+}
--- a/src/components/aeDivider/aeDivider.js
+++ b/src/components/aeDivider/aeDivider.js
@@ -1,6 +1,5 @@
 const aeDividerTypes = [
   'boring',
-  'normal',
   'exciting',
   'dramatic'
 ]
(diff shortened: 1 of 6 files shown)


src/components/aeDivider/aeDivider.md
```diff
@@ -0,0 +1,5 @@
+Divider
```
davidyuk: This shouldn't be necessary because `vue-styleguidist` already displays the name of a component.

src/index.js
```diff
@@ -81,6 +83,6 @@ export { default as AeSwitch } from './components/aeSwitch/aeSwitch.vue'
 export { default as AeAppIcon } from './components/aeAppIcon/aeAppIcon.vue'
 export { default as AeIcon } from './components/aeIcon/aeIcon.vue'
 export { default as AeAddress } from './components/aeAddress/aeAddress.vue'
-export { default as AeAddressInput } from './components/aeAddressInput/aeAddressInput.vue'
```
davidyuk: Why you have to remove AeAddressInput component?

src/components/aeDivider/aeDivider.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+  <div :class="cssClass">
+  </div>
```
davidyuk: I see that template for this component you have taken from `AeButton` or so. But I think better to write this component as simple as possible, so just `<div class="ae-divider" :class={ type } />` and remove `typeModifier` and `cssClass` computed variables.

src/components/aeDivider/aeDivider.vue
```diff
@@ -0,0 +1,7 @@
...
+</template>
+<script src='./aeDivider.js'/>
```
davidyuk: Is it difficult to use only one of `"` and `'` for quoting of a value of tag attribute?

src/components/aeDivider/aeDivider.js
```diff
@@ -0,0 +1,27 @@
+const aeDividerTypes = [
+  'boring',
+  'normal',
```
davidyuk: I propose to remove `normal` from divider types since it does nothing. Just apply corresponding background color by default to this component.

src/components/aeDivider/aeDivider.scss
```diff
@@ -0,0 +1,25 @@
+.ae-divider {
+  margin: 8px 0px;
+  box-sizing: content-box;
+  overflow: visible;
+  display: block;
```
davidyuk: `box-sizing`, `overflow`, `display` settings seems useless for this component.

src/index.js
Philip-Nunoo: :(

PR #35, opened 2018-01-26
Ae identity

Philip-Nunoo:
Requires aeDivider in this PR. https://github.com/aeternity/aepp-components/pull/34


src/components/aeIdentity/aeIdentity.vue
```diff
@@ -1,32 +1,42 @@
...
+        <!-- <div class="identity-info _long">{{address}}</div> -->
+        <div v-for="chunk in chunkAddress" class="chunk">
+          {{chunk}}
+        </div>
       </div>
-      <div class="bottom-inserted-elements" v-if="hasSlot && !collapsed">
-				<slot></slot>
+      <div class="footer" v-if="showButtons">
+        <ae-divider type="boring"/>
+				<div class="buttons">
```
davidyuk: Let's remove all tabs from this file.

src/components/aeIdentity/aeIdentity.vue
```diff
@@ -1,32 +1,42 @@
...
-        <div class="public-id-title">Public identity</div>
-        <div class="identity-info _long">{{address}}</div>
+      <div v-if="!collapsed" class="chunked-address">
+        <!-- <div class="identity-info _long">{{address}}</div> -->
+        <div v-for="chunk in chunkAddress" class="chunk">
+          {{chunk}}
+        </div>
       </div>
-      <div class="bottom-inserted-elements" v-if="hasSlot && !collapsed">
-				<slot></slot>
```
davidyuk: Will be better to pass buttons through slots, because there can be [other buttons set](https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5a5f16e1f080c170d9e73434), [qr code](https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5a3004ef20b1355a5469a7a7), and even [emptiness](https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5a27ffe6b1549395c783c17b).

src/components/aeIdentity/aeIdentity.js
```diff
@@ -8,7 +10,9 @@ import BN from 'bn.js'
 export default {
   name: 'ae-identity',
   components: {
-    'ae-identity-avatar': aeIdentityAvatar
+    'ae-identity-avatar': aeIdentityAvatar,
+    'ae-divider': aeDivider,
+    'ae-button': aeButton
```
davidyuk: Key can be omitted, so just
```
components: {
  aeIdentityAvatar,
  aeDivider,
  aeButton
}
```

src/components/aeIdentity/aeIdentity.vue
Philip-Nunoo: I actually thought about that. But was thinking it would actually easier to control both the design and bottom functionality from the component itself.

We can set display modes for each bottom if necessary.

davidyuk: No, it is not easier. For each bottom part, AeIdentity should define additional props and events.
For example, you added `toggleActive`, `edit`, `copyAddress` events, if you won't remove the slot in AeIdentity, a programmer must know just AeButton api.
The design of bottom functionality also can be controlled outside of AeIdentity.

src/components/variables.scss
```diff
@@ -12,3 +12,7 @@ $plum: #c69cd0;
 
 $container-width: 880px;
 $screen-phone: 480px;
+
+// new variables
+$exciting: #311b58;
+$plain: #ffffff;
```
dadaphl: - #311b58 is the updated value for $aubergine.
- please don't introduce color-variable named after purpose. the color/background of a `exciting` component is dependent on the context/contrast of a component. for example a exciting button can have a purple or white background depending on the background of the parent element. it would be confusing to write `.parent.exciting > .child.exciting { background : $plain}`

src/components/aeIdentity/aeIdentity.scss
```diff
@@ -0,0 +1,139 @@
...
+    justify-content: space-between;
+  }
+
+  .chunked-address {
+    display: flex;
+    flex-wrap: wrap;
+    margin-top: 10px;
+    margin-left: -10px;
+
+    .chunk {
```
dadaphl: it's hard to understand the context of the `.chunk` on a smaller screen
please don't indent fore than two times. better write 
```
  .chunked-address .chunk {
...
```

src/components/aeIdentity/aeIdentity.scss
```diff
@@ -0,0 +1,139 @@
...
+
+.ae-identity ,
+.ae-identity * {
+  transition:0.4s;
+}
+
+.ae-identity._active_no{
+  background-image: linear-gradient(to bottom, #aeaeae, #dcdcdc);
+}
+#app.home .ae-identity {
```
dadaphl: I know this was introduced in a commit before, but please remove this. this is not the right location for this style. i think this actually belongs to a specific application.
component should be atomic and not assume existence of certain classes in parent components. component position should be controlled by the parent. its just not clear for a developer that uses a common name like `.home` why it changes the position of all ae-identity.

src/components/aeIdentity/aeIdentity.scss
```diff
@@ -0,0 +1,139 @@
...
+    background-image: linear-gradient(119deg, #f22f70, #42255c);
+  }
+
+  &.collapsed{
+  }
+
+  .footer {
+    margin-top: 18px;
+
+    .buttons {
```
dadaphl: this class is never used in the component. we can not assume, that developers will use a class like that in the slot, and if they do, know what it means

src/components/aeIdentity/aeIdentity.scss
```diff
@@ -0,0 +1,139 @@
...
+  zoom: 1.1;
+  padding:15px 21px;
+  color:white;
+  z-index:5;
+  margin:0 auto;
+  text-align: left;
+  width: 335px;
+  box-shadow: 0 4px 8px 0 rgba(60, 60, 60, 0.1);
+  border-radius: 12px;
+  background-image: linear-gradient(122deg, #6f53a1, $exciting);
```
dadaphl: please change name of the color to $aubergine

src/components/aeIdentity/aeIdentity.js
```diff
@@ -7,12 +9,7 @@ import BN from 'bn.js'
  */
 export default {
   name: 'ae-identity',
-  components: {
-    'ae-identity-avatar': aeIdentityAvatar
-  },
-  data: function () {
-    return {}
-  },
+  components: { aeIdentityAvatar, aeDivider, aeButton },
```
dadaphl: aeButton is not used in the template, why was it added?

Philip-Nunoo:
Review made

src/components/aeIdentity/aeIdentity.vue
```diff
@@ -1,32 +1,39 @@
...
-            <span class="amount">{{amount}}</span>
-            <span class="currency-symbol">ETH</span>
+          <slot name="header-right"></slot>
+        </div>
+        <div v-if="!collapsed" class="chunked-address">
+          <div v-for="chunk in chunkAddress" class="chunk">
+            {{chunk}}
           </div>
         </div>
+        <div class="footer" v-if="slot">
```
dadaphl: this variable does not exist and the slot became inaccessible.

src/components/common.scss
```diff
@@ -1,4 +1,4 @@
-@import url('https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Roboto+Mono:500');
+@import url('https://fonts.googleapis.com/css?family=Roboto+Mono:100,300,400,500,700,900');
```
dadaphl: Again, I don't think we need to load all these fonts. Please check Zeplin Styleguide

src/components/aeIdentity/aeIdentity.js
```diff
@@ -67,18 +70,15 @@ export default {
...
         'ae-identity': true,
         'collapsed': this.collapsed,
-        '_active_yes': this.active,
-        '_active_no': !this.active
+        'active': this.active
       }
-      classes['size_' + this.size] = true
-      return classes
-    },
-    hasSlot () {
```
dadaphl: We need this for to access the slot. Why did you remove it?

src/components/aeIdentity/aeIdentity.js
```diff
@@ -40,20 +36,27 @@ export default {
...
     helperMixin
   ],
+  methods: {
+    /**
+     * Make card active event
+     *
+     * @event activate
+     * @type {undefined}
+     */
+    toggleActiveIdentityCard () {
```
dadaphl: When will this method be used?

src/components/aeIdentity/aeIdentity.js
```diff
@@ -40,20 +36,27 @@ export default {
...
-    * @deprecated use collapsed instead
+    * Have a blank identity card
     */
-    size: {
-      type: String,
-      default: 'small',
-      validator: function (size) {
-        return size === 'small' || size === 'big'
-      }
+    blank: {
```
dadaphl: What is a blank card and what do we need it for? I think this doesn't exist according to the designs.

src/components/aeIdentity/aeIdentity.js
```diff
@@ -30,7 +26,7 @@ export default {
     */
     active: {
       type: Boolean,
-      default: true
```
dadaphl: Why change it?

src/components/aeIdentity/aeIdentity.js
Philip-Nunoo: I created a blank card in order to satisfy this design.
image

src/components/common.scss
Philip-Nunoo: @dadaphl We are loading [RobotoMono](https://fonts.google.com/specimen/Roboto+Mono) only. It's one font being loaded here as per the styleguide

src/components/aeIdentity/aeIdentity.js
Philip-Nunoo: we are not using size anymore, but instead relying on the collapsed attribute.

src/components/aeIdentity/aeIdentity.js
Philip-Nunoo: I think at best within the app, most cards are supposed to be inactive except for one. So it would be best to set the default to *false* rather than true for every case.

Philip-Nunoo:
@dadaphl @davidyuk Since this has been recreated here https://github.com/aeternity/aepp-components/pull/42 I am closing this PR. And possibly request for the branch to be deleted as well.

davidyuk:
@Philip-Nunoo I suppose you wanted to close #27.

src/components/aeIdentity/aeIdentity.js
dadaphl: I won't accept changing this because of several reasons:

1) This change is not related to the rest of the changes in this PR. If we want to have a discussion about this, it becomes blurred and mixed with the rest of the changes. It blocks our progress. 

2) It's an arbitrary change that you decided to make based on a personal preference. It could potentially break others people code that don't expect such arbitrary change of default behaviour. Also it could inspire other devs to demand similar changes and distract what is really important.

3) I feel like `inactive` identity is the rarer (more special, thus requires more 'work') use case.
for example; html input element is also not 'disabled' by default.

4) Changing default behaviour is not creating any progress. It will not make anybodies experience with the components significantly easier to understand or faster to develop. 


I understand, this is just a tiny change and it might seem like I'm overly strict and conscious. However I believe that we have to start **training and enforcing a good workflow**, like separation of concerns, **right away**. As complexity grows rapidly, we would simply miss the moment, when it will be of crucial importance. Also our PRs should be a best practice example for other developers.

src/components/aeIdentity/aeIdentity.js
dadaphl: sorry, it was not clear. I was asking about the removal of `hasSlot()` function.

dadaphl:
I'm gonna close this for now because there is no progress or commitment for progress

PR #36, opened 2018-01-30
Aepp component doc

Philip-Nunoo:
New Aepp component documentation proposal. I propose we begin grouping the various components and into their respective groups and have a well-written documentation style for components.aepps.com. This is a sample documentatation proposal using styleguide.


docs/Introduction.md
```diff
@@ -0,0 +1,5 @@
+The aeternity aepp components are meant to provide aepp developers with reusable [Vue.js](https://vuejs.org) components that adhere to our [styleguide](https://github.com/aeternity/aepp-prototypes). Until the javascript comes bundled please use a build process such as [webpack](https://webpack.js.org/), [browserify](http://browserify.org/), [grunt](https://gruntjs.com/), [gulp](https://gulpjs.com/), etc.
+
+## Documentation
+
+Full documentation can be found at [http://components.aepps.com](http://components.aepps.com)
```
dadaphl: https

dadaphl:
I think its a very good idea to improve the documentation. I don't understand though what you mean with `grouping the various components and into their respective groups`. Would that mean, that we make up groups and associate the components to this groups? Do this groups already exist somehow and I'm not aware of it? Do you have a proposal for this specific groups? Would the documentation of a component stay inside the `src/components/$name/$name.md` file? 
I think you need to write more examples to illustrate the idea better. So far its not clear for me what's the actual goal.

Philip-Nunoo:
@dadaphl "

> Would that mean, that we makeup groups and associate the components to this groups, 

yes but that's just my proposal. Like everything with respect to forms is grouped as one. But it's only a proposal from my end.

davidyuk:
I understood the idea. So, let's define some groups and merge this PR.

dadaphl:
@Philip-Nunoo I think you should just propose this groups and create more commits.

dadaphl:
i'm gonna close this for now because there is no progress or commitment for progress

PR #42, opened 2018-02-03
Update AeBanner component

davidyuk:
The rebased version of #27 with some improvements.

screenshot from 2018-02-04 00-02-40


src/components/aeBanner/aeBanner.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.ae-banner {
+  background: linear-gradient(to bottom, rgba($aubergine, 0.65), transparent);
+  z-index: 1;
+  padding: 10px 10px 25px 10px;
+
+  > div {
+    display: flex;
+    align-items: center;
+    height: 64px;
```
Philip-Nunoo: I think at best having a padding instead of a fixed height would be ok. 
`
padding: 17px 22px;
`

dadaphl: i agree

PR #44, opened 2018-02-05
Update AeAddressInput component

davidyuk:
screenshot from 2018-02-05 16-11-18
Component required in [this mockup](https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5a27ffe7920505c5488cbe1f).

Workaround for Chrome on Android is based on https://github.com/text-mask/text-mask/pull/400.

Also, I checked most popular mask packages for vue:
- [`vue-text-mask`](https://www.npmjs.com/package/vue-text-mask) creates [only `<input />` element](https://github.com/text-mask/text-mask/blob/master/vue/src/vueTextMask.js)
- [`vue-the-mask`](https://www.npmjs.com/package/vue-the-mask) `mask` directive [works only with `<input />` element](https://github.com/vuejs-tips/vue-the-mask/blob/master/src/directive.js#L23)
- [`v-mask`](https://www.npmjs.com/package/v-mask) requires [`v-model` usage](https://www.npmjs.com/package/v-mask#usage)
- [`vue-masked-input`](https://www.npmjs.com/package/vue-masked-input) has [bugs in mobile Chrome](https://www.npmjs.com/package/vue-masked-input#known-issuestodo)


src/components/aeInput/aeInput.vue
```diff
@@ -36,13 +41,72 @@
...
+            : address.slice(0, fl)
+          while (group() !== '') {
+            res.push(group())
+            i++
+          }
+          return [res.join(' '), res.length ? res[res.length - 1].length : 0]
+        }
+
+        let lastLength
+        [begin, lastLength] = addSpaces(begin);
```
davidyuk: I have found a case when a semicolon at end of line in JS is necessary.

dadaphl:
its a very well build component and matches the design. However we already have a `ae-address-input` component. can you please have a look, if we still need it? I think it is used in `aepp-identity`. 

I would  like to:
- **either** get rid of `ae-address-input`
- **or** port `ae-input[type:address]` to `ae-address-input

Also:
ae-input could be a convenience wrapper for complex ae-...-input components.

@davidyuk @Philip-Nunoo which one do you think is better? I think this is a turning point in our general component structure design.

- Do we want to have convenience wrappers?
- Do we want to have thin or fat components?

I, as a aepp developer, want to keep my app as light as possible. If i have a ae-input that supports 10 use-cases and I only need 2, I'd feel that I'm installing overhead. So:

I'm am favouring **thin** components and **no convenience** wrappers

acdbaykal:
Small is always better. Besides the type attribute for input elements has proven itself to be a mess. There are currently 17 types. Some other attributes of input are only useful when the input has a certain type (like max and min for the types date and number). The different types look and behave differently. I.e. `date` opens a calendar and  `number ` provides up and down arrows. We should avoid going down the same path. 

I think `ae-address-input` should lose its labels and validation messages. The labels and validation messages should move to a new component, which takes an input as a slot.

Additionally `ae-input` could gain an optional `formatText` function as a prop. Than `ae-address-input` coud simply be more concrete variant of `ae-input`

dadaphl:
Ok lets improve `ae-address-input` with @davidyuk 's code and solve the label problem with similar approach than #25 

I'm against a 'formatText' prop in ae-input, this component became already complex enough for now.

davidyuk:
So, AeAddressInput should have a label?

acdbaykal:
@davidyuk, no, the component should not have a label

PR #46, opened 2018-02-08
 Update AeAmountInput component

davidyuk:
screenshot from 2018-02-08 19-39-10


src/components/aeAmountInput/aeAmountInput.md
```diff
@@ -1,18 +1,19 @@
...
-})
+```js
+  new Vue({
+    data: () => ({ value: undefined }),
+    template: `
+      <div>
+        <ae-amount-input
+          placeholder="0.00"
+          v-model="value"
+          :tokens="[
```
dadaphl: I think we should change the name to something more abstract. What if i want to use it to input lengths, 
```
:tokens="[{ symbol: 'CM', name: 'centimeters' },{ symbol: 'IN', name: 'inches' }]
```
maybe we better call it "unit"?

src/assets/icons/chevron-down.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><path transform="rotate(90 12 12)" d="M17.192 11.778L9.414 4 8 5.414l6.364 6.364L8 18.142l1.414 1.414z"/></svg>
```
dadaphl: as this is only a rotated version of chevron. It would be nice if instead of using a new file, we could come up with a good solution to reuse files we already have
see #49

src/components/aeAmountInput/aeAmountInput.md
dadaphl: see #50

Issue #52, opened 2018-02-14
AeLabel without margin and without display:flex


dadaphl:
AeLabel should not have a margin and make assumptions about how it is displayed. It makes it difficult to display and position labels properly in an application

davidyuk:
In which case you need different margins of AeLabel? What wrong with displaying in flex mode?

PR #56, opened 2018-02-16
AeLabel: Make "help-type" the same as "type" of "AeButton"

davidyuk:
#52


acdbaykal:
@dadaphl, improvements can be found in #62

PR #59, opened 2018-02-19
AeIcon: Add `rotate`, `size` props

davidyuk:
screenshot from 2018-02-19 17-29-17

Though small icons are thinner than existing one:
screenshot from 2018-02-19 17-24-03


acdbaykal:
:+1:

dadaphl:
I thought that 'error' really was indeed a different icon? I'd appreciate if PRs are more atomic. Please for the future, create a PR that does one thing at a time. In this case, implement the rotation. Deleting unused icons should be a different PR so we can have a discussion about it without blocking.

davidyuk:
@dadaphl  Ok, I have brought 'error' icon back since it is bolder than small 'close' icon.

PR #62, opened 2018-02-20
AeLabel help-type supporst all types (change suggestion to #56)

acdbaykal:
This version of #56 
- supports all types ('boring', 'normal', 'exciting', 'dramatic')
- makes 'normal' the default type (which it implicitly already was)
- follows our CSS modifier class name convention

### DIFF
--- a/src/components/aeButton/aeButton.js
+++ b/src/components/aeButton/aeButton.js
@@ -1,11 +1,6 @@
 import AeLink from '../aeLink/aeLink.vue'
+import {TYPE_PROPERTY_VALUES as aeButtonTypes} from '../../constants'
 
-const aeButtonTypes = [
-  'boring',
-  'normal',
-  'exciting',
-  'dramatic'
-]
 const aeButtonSizes = [
   'smaller',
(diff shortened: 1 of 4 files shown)


davidyuk:
Can I find somewhere description of our CSS modifier class name convention? I think this convention if it exists is too redundant, let's keep things simple.
Explicitly defined `normal` type is useless.
I'm not sure that `exciting` help type will be used somewhere, not to mention `boring`.
Despite this, I like the idea to generate color styles using cycles.

acdbaykal:
Hi @davidyuk,
https://github.com/aeternity/aepp-components#component-css-classes

Putting some underscores in class names isn't complicated and imo it does communicate the purpose of the css class better. 

It is hard to predict, how these components will be used by others. It isn't entirely out of question that some one may decide to use those types. However my main point would be, that consistency is important for us human beings. Since we have a naming/coloring theme we should implement that every where, fully. And I think #52 mirrors notion, since renaming the type from "Danger" was not requested out of technical inadequacy but consistency.

PR #63, opened 2018-02-21
fix squash image when right header is added

Philip-Nunoo:
fixes image being squashed when right header is added.
image

image


acdbaykal:
Manually tested the solution with latest Ff and Chrome, IE 15, Android Chrome,  X OS Safari 10 and Safari on iPhone 6.

acdbaykal:
I allowed myself to change the commit message to: "AeIdentity: Fixed squashed identity avatar when right header is added"

PR #67, opened 2018-03-02
AeIdentity: Fix default value of 'identity' prop

davidyuk:
[Vue warn]: Invalid default value for prop "identity": Props with type Object/Array must use a factory function to return the default value.

found in

---> &lt;AeIdentity&gt;


acdbaykal:
I tested this with the style guide page. I was able confirm the fix.

Issue #69, opened 2018-03-10
Modify AeAddress validation to accept AE addresses


tillkolter:
I integrated the AeAddress field for the [æpp showcases](https://github.com/aeternity/aepp-showcases/tree/feature/faucet_initialisation) and found a validation warning for the æternity addresses.

![screen shot 2018-03-10 at 17 53 49](https://user-images.githubusercontent.com/590062/37244627-0f7f5a5e-248c-11e8-8cfe-75f91bd68358.png)

tillkolter:
`ak$3Q1hfV4j3HdKHyugyDSTsK9pm1nvBK3q72UYQ41UuRJ7ei1QWskoEPZzpzgk5WJCi4efcRrCvoLHMrE9PBG1WZ1zL8q5Bz`

PR #70, opened 2018-03-12
Fix issue #69/AeAddress: Accept ak$ and ok$ addresses

acdbaykal:
Extended the validation regex to accept addresses with a length of 97 characters and starting with ak$ or ok$.


dadaphl:
aeternity addresses are base58 encoded, as far as i understand the length of the addresses are not always the same. please check and confirm if i'm wrong

acdbaykal:
@dadaphl, I did not find any official documentation on the matter so I asked the core team. I got following answer from @sennui 

> OK, true, so the input has constant length, hence the base58 output will keep it

Issue #72, opened 2018-03-19
Identity component broke on mobile breakpoint


SamLife91:
- prevent text overfloating
 - fix  ae-buttons layout in identity when mobile breakpoint is active

PR #73, opened 2018-03-19
[GH-#72] indentity component mobile breakpoint fix

SamLife91:
- remove overflowing text content
 - format code (nesting, replace hd-coded colors with ther vars)


dadaphl:
as we have discussed, i will close this PR now because the bug only occurs in the documentation .

Issue #74, opened 2018-03-19
AeInputAmount should not display chevron when only one unit is present


dadaphl:
The symbol should not be clickable. There should be no drop-down
```
    <ae-amount-input
      placeholder="0.00"
      v-model="value"
      :units="{ symbol: 'AE', name: 'æternity' }"
    />
```

# currently
<img width="549" alt="screen shot 2018-03-19 at 23 39 44" src="https://user-images.githubusercontent.com/415536/37626135-f972ba68-2bce-11e8-92b8-0daa75ce1336.png">

# expected
![screen shot 2018-03-19 at 23 39 44](https://user-images.githubusercontent.com/415536/37626182-2191e334-2bcf-11e8-8983-19d8e1d7db7e.png)

Issue #75, opened 2018-03-19
AeInputAmount default display unit should be generated from units


dadaphl:
Default unit is AE even if AE doesn't exist in units prop. 

# current
<img width="523" alt="screen shot 2018-03-19 at 23 53 34" src="https://user-images.githubusercontent.com/415536/37626555-c84be458-2bd0-11e8-8057-257c1395eb5f.png">

# expected
<img width="524" alt="screen shot 2018-03-19 at 23 53 13" src="https://user-images.githubusercontent.com/415536/37626558-cbc733e4-2bd0-11e8-985a-ab3fae7af56a.png">

PR #76, opened 2018-03-21
Hide chevron if only one unit is passed

SamLife91:
- this PR resolve issue #74 and issue #75

### DIFF
--- a/src/components/aeAmountInput/aeAmountInput.vue
+++ b/src/components/aeAmountInput/aeAmountInput.vue
@@ -47,7 +47,7 @@
        */
       value: {
         type: Object,
-        default: () => ({ symbol: this.valueComputed})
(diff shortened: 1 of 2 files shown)


src/components/aeAmountInput/aeAmountInput.vue
```diff
@@ -62,14 +62,24 @@
...
+      },
+      toggleDropDown () {
+        if (this.unitsCount > 1) {
+          this.dropDownVisible = !this.dropDownVisible
+        }
       }
+    },
+    mounted () {
+      this.value.symbol = this.units[0].symbol
+      this.unitsCount = this.units.length
```
dadaphl: this could be done with a computed property, so if you have unit set dynamic it would also stay up to date.

PR #84, opened 2018-03-26
aeAccountBackground: A component as background for identity related elements

acdbaykal:
[](https://postimg.org/image/p2vj4dg2j/)

### DIFF
--- a/src/components/aeAccountBackground/aeAccountBackground.js
+++ b/src/components/aeAccountBackground/aeAccountBackground.js
@@ -1,0 +1,23 @@
+import {TYPE_PROPERTY_VALUES as types} from '../../constants'
+
+export default {
+  name: 'ae-account-background',
+  props: {
+    type: {
+      type: String,
+      default: 'normal',
+      validator: value => types.includes(value)
+    }
+  },
+  computed: {
+    typeModifier () {
+      return `_type_${this.type}`
+    },
+    cssClasses () {
+      return [
+        'ae-account-background',
+        this.typeModifier
+      ]
+    }
+  }
+}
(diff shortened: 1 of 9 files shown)


davidyuk:
AePanel, AeBanner, and AeButton technically is also a rounded container. Maybe here can be used a more particular name? Like AeCard or AeAccountBackground.

src/components/aeRoundedContainer/index.js
```diff
@@ -0,0 +1,5 @@
+import aeIdentityAvatar from './aeRoundedContainer.vue'
```
davidyuk: `aeIdentityAvatar` should be replaced with `aeRoundedContainer`

acdbaykal: Whoops show did that slip in :(

acdbaykal:
I like AeAccountBackground. It's not always used like a card.

src/components/aeRoundedContainer/aeRoundedContainer.js
```diff
@@ -0,0 +1,54 @@
...
+      default: 'stretch',
+      validator: value => ['start', 'end', 'center', 'baseline', 'stretch'].includes(value)
+    },
+    alignContent: {
+      type: String,
+      default: 'stretch',
+      validator: value => [
+        'start', 'end', 'center', 'space-between', 'space-around', 'stretch'
+      ].includes(value)
+    }
```
davidyuk: Even in `scoped` mode, it is possible to set styles of the root node of child components, Is it necessary to wrap possible css properties with vue component properties with so lot of code?

src/components/aeRoundedContainer/aeRoundedContainer.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <span :class="cssClasses">
```
davidyuk: I think `div` has a better semantic for something with `display: flex`.

PR #85, opened 2018-03-27
AeIdentity fixes

davidyuk:
Perhaps `this.$slots.default` is not a reactive variable since `hasSlot` contains an old value of `$slots.default` after its update.
@Philip-Nunoo is there any reason to enable `transition` for the root node and all children?


acdbaykal:
It allows buttons to be added dynamically (without resizing failure) and increases reusability by removing transitions.

👍

PR #87, opened 2018-04-03
ae-address redesign

SamLife91:
- change design of aeAdddress component
- add addition unit tests

### DIFF
--- a/src/components/aeAddress/aeAddress.vue
+++ b/src/components/aeAddress/aeAddress.vue
@@ -1,12 +1,18 @@
 <template>
-  <div :class='cssClass'>
+  <div :class='cssClass' class="aeAddressWrapper">
     <ae-identity-avatar v-if='showAvatar' 
     :address='address' 
     :name='name'
+    class="avatar"
     />
     <div>
-      <span class='name'>{{displayName}}</span>
-      <span class='address' v-if="size !== 'full'">{{displayAddress}}</span>
+      <div class='name'>
+        {{displayName}}
+        <ae-icon name='check'
+            type='dramatic'
+            v-if='verified' />
+      </div>
+      <div class='address' v-if="size !== 'full'">{{displayAddress}}</div>
       <div v-if="size === 'full'" class="chunked-address">
           <div v-for="chunk of displayAddress" class="chunk">
               {{chunk}}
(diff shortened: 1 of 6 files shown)


acdbaykal:
Should the version below also be supported?

[](https://postimg.org/image/5g4266lqf/)

acdbaykal:
Same question for the next one

[](https://postimg.org/image/lfmppqphz/)

acdbaykal:
Than there is also one with dots at the instead of the middle

[](https://postimg.org/image/wghuuohpz/)

src/components/aeAddress/aeAddress.vue
```diff
@@ -1,9 +1,26 @@
 <template>
-  <div :class='cssClass'>
-    <ae-identity-avatar v-if='showAvatar' :address='address' />
-    <span class='address'>{{displayAddress}}</span>
+  <div :class='cssClass' class="aeAddressWrapper">
```
acdbaykal: I'm really not a fan of using :class and class on the same element. I prefer something like

```
:class="['aeAddressWrapper', cssClass]"
```

src/components/aeAddress/aeAddress.vue
```diff
@@ -1,9 +1,26 @@
...
+          {{chunk}}
+        </div>
+      </div>
+    </div>  
   </div>
 </template>
 <script src='./aeAddress.js'/>
 /* eslint no-unused-expressions: "off" */
-<style src='./aeAddress.scss'/>
+<style scoped="scoped" src='./aeAddress.scss'/>
```
acdbaykal: Good you cached that!

```
<style scoped src='./aeAddress.scss'/>
```
is sufficient though

dadaphl:
@raymundjacobs 
Is it true, that we need a AeAddress with the ellipsis `...` in the end? Maybe we can avoid this and rethink this specific screen.

dadaphl:
I suggest that the AeAddress component stays simple by default. The new layout should be a variation.

src/components/aeAddress/aeAddress.vue
acdbaykal: No, not that I know. Maybe Vue is slightly more performant one variation Idk. My reason is that I like to have it in one place.

SamLife91:
@acdbaykal @dadaphl @raymundjacobs  i can implement this as additional (optional) layout or type

PR #91, opened 2018-04-04
Fix ae amount

acdbaykal:
Related to https://github.com/aeternity/aepp-identity/issues/93 and https://github.com/aeternity/aepp-identity/pull/95


dadaphl:
size `med` should be `medium` to be consistent with button size.

PR #92, opened 2018-04-05
ae-address redesign: Altered unit tests

acdbaykal:
@SamLife91, 

I don't think testing for class names is that important. It's more about functional stuff like which components are rendered when, how are events handled and so on.


acdbaykal:
Merged manually

PR #94, opened 2018-04-06
Update aeIdentity

acdbaykal:
- This PR's main intent is to update the aeIdentity component to the enhanced design.
- It also extracts the background and main content of aeIdentity in to two separate components
- It simplifies aeIdentity by removing the header-left and header-right slots

The new design adds a name and a smaller sized address into the header. And also requires a collapsed state to have buttons on the bottom.

screenshot-2018-4-20 aeternity identity

Two components out of aeIdentity in order to make building component like the following easier in the future.
screenshot-2018-4-20 zeplin - project 1

screenshot-2018-4-20 zeplin - project 2

header-left and header-right slots were removed because they are only used within QuickId in the Identity Manager. Imo they were cluttering the component. I suggest re-implementing QuickId using the new aeAccountBackground and aeIdentityMain components, as I did in https://github.com/aeternity/aepp-identity/pull/103 .


src/components/aeAccountBackground/aeAccountBackground.md
```diff
@@ -0,0 +1,32 @@
...
+</ae-account-background>
+```
+
+```js
+const BN = require('bn.js')
+let identity = {
+  name: 'My identity',
+	address: '0x35d8830ea35e6Df033eEdb6d5045334A4e34f9f9',
+	balance: new BN('1337000000000000000', 10)
+}
```
davidyuk: Here and in the example below are missed space in indention. Also, I prefer to use `const` instead of `let`, if a variable is not going to be reassigned.

src/components/aeAccountBackground/aeAccountBackground.scss
```diff
@@ -0,0 +1,20 @@
...
+    @each $type in map_keys($type-color-map) {
+      &_#{$type} {
+        $color: #{map_get($type-color-map, $type)};
+        background-color: $color;
+        border-color: $color;
+      }
+    }
+  }
+}
+
```
davidyuk: Extra newline at the end of file.

src/components/aeIdentityAvatar/aeIdentityAvatar.vue
```diff
@@ -2,6 +2,8 @@
   <div :style='style' class="avatar ae-identity-avatar"></div>
 </template>
 
-<script src='./aeIdentityAvatar.js'/>
+<script src='./aeIdentityAvatar.js'>
+  export { default } from './aeIdentityAvatar'
```
davidyuk: Is it necessary to add this?

src/components/aeIdentityMain/aeIdentityMain.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div @click="forwardEvent" :class="classObject">
```
davidyuk: Is `forwardEvent` defined somewhere?

src/components/aeIdentityMain/aeIdentityMain.scss
```diff
@@ -0,0 +1,125 @@
+@import './../variables';
+
+.ae-identity {
+  width: 100%;
+  zoom: 1.1;
```
davidyuk: Looks like a workaround, maybe better to explicitly make font sizes and avatar bigger?

src/components/aeIdentityMain/aeIdentityMain.scss
```diff
@@ -0,0 +1,125 @@
...
+
+.currency-symbol {
+  letter-spacing: 0.12em;
+  padding-right: 3px;
+}
+
+.ae-identity__avatar {
+  border-width: 1px !important;
+  width: 23px !important;
+  height: 23px !important;
```
davidyuk: Please, make selector more specific to avoid `!important`.

src/components/aeIdentityMain/aeIdentityMain.vue
```diff
@@ -0,0 +1,36 @@
...
+        <header :class="['identity-name', collapsedModifier]">{{identityName}}</header>
+        <small class="truncated-address" v-if="collapsed" data-truncated-address>{{shortAddress}}  ••••••</small>
+      </span>
+      <div class="balances">
+        <div v-if='tokenAmount' class="balance _bold">
+          <span class="amount" data-token-amount>{{tokenAmount}}</span>
+          <span class="currency-symbol">AE</span>
+        </div>
+        <div class="balance">
+          <span class="amount" data-amount>{{amount}}</span>
```
davidyuk: `data-amount`, `data-token-amount`, `data-truncated-address` it is just for tests? You shouldn't write here anything for tests.

src/components/aeIdentityMain/aeIdentityMain.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div @click="forwardEvent" :class="classObject">
+    <div class="flex-row">
+      <ae-identity-avatar
+        class="ae-identity__avatar"
+        :address='identity ? identity.address : ""'
+        @click="forwardEvent"
+      />
+      <span :class="['identity-name-position', collapsedModifier]">
+        <header :class="['identity-name', collapsedModifier]">{{identityName}}</header>
```
davidyuk: According to https://validator.w3.org/:
```
Element header not allowed as child of element span
Contexts in which element header may be used: Where flow content is expected.
Content model for element span: Phrasing content.
```

src/components/aeIdentityMain/aeIdentityMain.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div @click="forwardEvent" :class="classObject">
+    <div class="flex-row">
+      <ae-identity-avatar
+        class="ae-identity__avatar"
```
davidyuk: This component already has `ae-identity-avatar` class, it is not necessary to define the similar class.

src/components/aeIdentityMain/aeIdentityMain.scss
```diff
@@ -0,0 +1,125 @@
...
+.identity-name-position {
+  &._collapsed {
+    &_true {
+      margin-left: 9px;
+    }
+    &_false {
+      margin-left: 10px;
+    }
+  }
+}
```
davidyuk: `_collapsed_true/false` makes code too verbose, why don't define just `_collapsed` and override default style with it?
```
.identity-name-position {
  margin-left: 10px;
  &._collapsed {
    margin-left: 9px;
  }
}
```

src/components/aeIdentityMain/aeIdentityMain.js
```diff
@@ -0,0 +1,64 @@
...
+    helperMixin
+  ],
+  computed: {
+    amount () {
+      return this.identity ? helperMixin.methods.readableEther(this.identity.balance) : 0
+    },
+    tokenAmount () {
+      return this.identity && this.identity.tokenBalance ? helperMixin.methods.readableToken(this.identity.tokenBalance) : '0'
+    },
+    address () {
```
davidyuk: This computed variable is unused.

src/components/aeIdentityMain/aeIdentityMain.js
```diff
@@ -0,0 +1,64 @@
...
+      return this.identity.address
+    },
+    shortAddress () {
+      return this.identity.address.substr(0, 8)
+    },
+    identityName () {
```
davidyuk: I think this alias is useless, it is used only in the template and shorter only by one character than `identity.name`.

src/components/aeIdentityMain/aeIdentityMain.js
```diff
@@ -0,0 +1,64 @@
...
+    identity: {
+      type: Object,
+      default: () => ({
+        address: '0x0',
+        tokenBalance: new BN('0', 10),
+        balance: new BN('0', 10)
+      })
+    },
+    /**
+    * Is this the full size identity card or the collapsed one for displaying at the bottom of the screen?
```
davidyuk: The collapsed identity displayed not only at the bottom of the screen but in identities list, probably in the address book and so.

src/components/aeIdentity/aeIdentity.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+      const wrapper = shallow(AeIdentity, {
+        stubs: {
+          AeIdentityMain
+        }
+      })
+
+      wrapper.setProps({ identity })
+
+      return wrapper.vm.$nextTick().then(() => {
```
davidyuk: I think to use async/await here will be more clear.

src/components/aeIdentity/aeIdentity.test.js
```diff
@@ -0,0 +1,99 @@
...
+        return wrapper.vm.$nextTick().then(() => {
+          const main = wrapper.find(AeIdentityMain)
+          expect(main.vm.$props.collapsed).toBe(collapsed)
+        })
+      }
+
+      return Promise.all([
+        test(true),
+        test(false)
+      ])
```
davidyuk: Running tests in parallel can be more difficult to debug. I propose:
```
await test(true)
await test(false)
```

src/components/aeIdentity/aeIdentity.md
```diff
@@ -69,14 +54,35 @@ let identity = {
...
 }
-<ae-identity :identity="identity" collapsed showButtons />
+<ae-identity :identity="identity" collapsed >
+   <ae-divider type="boring"/>
+    <div>
+      <ae-button type="boring" size="small" uppercase>Edit</ae-button>
+      <ae-button type="boring" size="small" uppercase>Copy</ae-button>
+      <ae-button type="dramatic" size="small" uppercase :inactive="active">
+        <span v-if="active">Active</span>
+        <span v-else>Activate</span>
```
davidyuk: For what reason conditional rendering here if `active` cannot be changed?

src/components/aeIdentity/aeIdentity.md
```diff
@@ -69,14 +54,35 @@ let identity = {
...
+      </ae-button>
+    </div>
```
davidyuk: This part has wrong indents.

src/components/aeIdentity/aeIdentity.md
```diff
@@ -18,37 +18,22 @@ let identity = {
...
-}
-
-<ae-identity :identity="identity" active collapsed>
-  <ae-icon name="chevron" type="plain" slot="header-right"/>
-</ae-identity>
-
+<ae-identity :identity="identity" active />
 ```
 
+Named basic Identity card
```
davidyuk: This example is the same as the first one.

davidyuk:
Let's rename `AeIdentity*` components to `AeAccount*`

dadaphl:
The idea of an account is too abstract. What picture comes to your mind if you think about an account? Maybe `AeIdentity` is wrong, but i think `AeAccount` is not an improvement.

src/components/aeIdentity/aeIdentity.vue
```diff
@@ -1,33 +1,14 @@
...
-          {{chunk}}
-        </div>
-      </div>
-      <div class="footer" v-if="hasSlot">
-        <slot></slot>
-      </div>
+  <ae-account-background
+    :type="backgroundType"
+    class="ae-identity"
+    @click="$emit('click', $event)"
```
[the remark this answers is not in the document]
acdbaykal: I would prefer that too. Unfortunately it breaks the unit test :-\ I don't  know why.

src/components/aeIdentity/aeIdentity.vue
```diff
@@ -1,33 +1,14 @@
...
+  >
+    <ae-identity-main :identity="identity" :collapsed="collapsed" @click.stop="$emit('click', $event)"/>
```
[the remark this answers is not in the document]
acdbaykal: Yes, I tried it without before and the clicking the ae-identity-main did not work.

src/components/aeIdentityAvatar/aeIdentityAvatar.vue
acdbaykal: Yes, the unit tests don't import it right if you don't have it (sadly).

src/components/aeIdentityMain/aeIdentityMain.js
acdbaykal: A question comment :D I'll remove that

src/components/aeIdentityMain/aeIdentityMain.vue
acdbaykal: It's better if a component doesn't need to know anything about the internals of another component. More concretely ae-identity-main does not know what classes are used in ae-identity-avatar.

src/components/aeIdentityMain/aeIdentityMain.vue
acdbaykal: Well you need to select these elements and check their values somehow. If I use the class names as selectors than changing them will break the tests unnecessarily.   Adding something for better testability is better than not  having tests, right?

davidyuk:
`account` is a well-defined entity in our application, I don't see any problem to name these components in that way. Also, it is very inconsistent that in this PR we have both AeIdenitty* and AeAccountBackground components. We can make this decision later, in this case, AeAccountBackground should be renamed to AeIdentityBackground.

src/components/aeIdentityMain/aeIdentityMain.vue
davidyuk: You should use class names instead. It is okay to update tests if you make some changes to the corresponding component.

src/components/aeIdentityMain/aeIdentityMain.vue
davidyuk: Hm, we really have this convention:
https://github.com/aeternity/aepp-components/blob/37e01ee/README.md#component-css-classes

PR #102, opened 2018-04-24
Add icon copy.svg

dadaphl:
screen shot 2018-04-24 at 13 12 48


dadaphl:
but then the other icon would have to be called `two-squares`. i guess we should go with what they represent, i think eyes represent view/looking/details

davidyuk:
Corresponding icons in Fontawesome has name [eye](https://fontawesome.com/icons/eye) and [copy](https://fontawesome.com/icons/copy) or [clone](https://fontawesome.com/icons/clone).

PR #104, opened 2018-04-25
Enable more strict eslint vue rules

davidyuk:
Please, read this code by commits, I tried to make them very understandable.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,7 +23,6 @@ module.exports = {
       'vue': 'never'
     }],
     // allow debugger during development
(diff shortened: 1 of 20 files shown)


dadaphl:
title says "Enable more strict eslint vue rules" but there are lots of changes that seem to be unrealated. is that a mistake? 
What is "more strict"?

davidyuk:
Yes, these changes were made to pass linter with new rules.

davidyuk:
Except for 8ff2ce4 and ada63b2, I decided to make them while fixing of AeTextarea.

PR #120, opened 2018-06-28
[WIP] New component designs and styleguide: https://sketch.cloud/s/3nq5w

sqevani:

PR #123, opened 2018-08-08
Add AeToolbar component

davidyuk:
Based on #120


davidyuk:
@sadiqevani what's wrong with this particular PR?

At the last sprint planning as I understand we decided to apply my approach firstly. Maybe the next time you won't do a such big PRs.

sqevani:
I'm listing the points to as why this is a bad idea:

- Each component with its own PR, there are probably around +10/15 components currently
  - Let say its going to take you from 1 hours to 2 hours on creating those pull-requests, 20 hours for uselessly trying to do things just because you "don't" like how things are being done, because you haven't understood yet the scope of the components and the structure in place.
- You want to explicitly include `scss` global variables on each component SCSS, bad idea in general. too much redundant declaration.
  - Less code, the better!
  - Cumbersome to add those files every time you make a component
  - There is a need for a general consistency, and those global assets achieve that
  - You generally don't have to think of the global files. Only in case you want to edit/add
  - There is a build system which takes care of every single use-case in terms of exporting the final build
    - Everything in a single File (a `.css` and a `.js` file)
    - Separate components with a "global `.css` file" that applies general css rules
  - There is a specific structure to the `scss` files and folders that allow that level of flexibilit.
- Adding everything in a single `Vue` component, thats great for small apps in general, but for longterm big projects its better if different languages are in their own separate files, it reduces noise in the code. Also different editors will pick the right syntax, highlight, and so-on-and-so-fourth.
- Code consistency is also something you should think of. Don't write cryptic functions with the newest `hacks` you've found online. 
- For me generally the updates i've done on the #120 help me think clearly of the place I'm at, what I'm coding, what should be done.
- Also keep in mind about the build system, the size of the bundle went from something hundreds kb to 20kb total.
- Things need structure, thats why you can read these bullet-point items with ease, but imagine if I wrote everything in a single paragraph:

> Each component with its own PR, there are probably around +10/15 components currently. Let say its going to take you from 1 hours to 2 hours on creating those pull-requests, 20 hours for uselessly trying to do things just because you "don't" like how things are being done, because you haven't understood yet the scope of the components and the structure in place. You want to explicitly include `scss` global variables on each component SCSS. Less Code, the better! Cumbersome to add those files every time you make a component. There is a need for a general consistency, and those global assets achieve that. You generally don't have to think of the global files. Only in case you want to edit/add. There is a build system which takes care of every single use-case in terms of exporting the final build. Everything in a single File (a `.css` and a `.js` file). Separate components with a "global `.css` file" that applies general css rules. There is a specific structure to the `scss` files and folders that allow that level of flexibilit. Adding everything in a single `Vue` component, thats great for small apps in general, but for longterm big projects its better if different languages are in their own separate files, it reduces noise in the code. Code consistency is also something you should think of. Don't write cryptic functions with the newest `hacks` you've found online.  For me generally the updates i've done on the #120 help me think clearly of the place I'm at, what I'm coding, what should be done. Things need structure, thats why you can read these bulletpoint items with ease, but imagine if I wrote everything in a single paragraph, It will be difficult

davidyuk:
> Adding everything in a single Vue component, thats great for small apps in general, but for longterm big projects

is was already rejected in https://github.com/aeternity/aepp-blockchain-explorer/pull/29

davidyuk:
> Cumbersome to add those files every time you make a component

But it makes possible to track dependencies between components and styles.

davidyuk:
> Also keep in mind about the build system, the size of the bundle went from something hundreds kb to 20kb total

This changes should not be in the same PR with new components. Your code doesn't become better because of this.

davidyuk:
> Let say its going to take you from 1 hours to 2 hours on creating those pull-requests

You were able to do it at the beginning, why you decided to put all these things in single PR? Do not you realize that this is a bad practice after you did the same in https://github.com/aeternity/aepp-blockchain-explorer/pull/28?

sqevani:
@davidyuk 

I think if you merge this, you might include unnecessary CSS and other stuff into the master/v2.

davidyuk:
I'm agreed to continue to work with #120 merged into `develop` branch.

PR #139, opened 2018-08-29
Drop example code

davidyuk:


sadiqevani:
The `core/` folder is going to be removed. and be replaced by the 3 folders that you removed in there. So this pull-request is useless.

Also, please focus on working on other issues that are open right now, as you can see, i'm currently tackling this section.

sqevani:
1. This pull-request does multiple things
  - Removes example code
  - Renames other things uselessly
  - Generalises the mixins `events -> helpers` which its a bad idea to include a bundle of all mixins in a single file, as specific components have specific needs
  - I'm going to be dropping some stuff for the future, constants and some other things that are not needed.

So in general I'll handle the moving of example code for:
  - Events
  - Mixins
  - Components

to a folder called `contributions` where .md files are going to be defining contributions rules for specific areas for the components.

davidyuk:
> Renames other things uselessly

You have made actually the same in #120 and you have don't prove that it needs to do, why now I should prove it? This PR reverts the state of the master branch, it is enough to merge this PR.

> Generalises the mixins events -> helpers which its a bad idea to include a bundle of all mixins in a single file, as specific components have specific needs

I would like to merge `master` into `develop` branch. In the master branch, `helper` mixin contains only `forwardEvent` method.

PR #140, opened 2018-08-29
Drop webpack aliases

davidyuk:
In addition to arguments in [the similar PR](https://github.com/aeternity/aepp-identity/pull/142) I would like to add that custom aliases makes mode difficult to build components package as a dependency of another project, when it is imported by `@aeternity/aepp-components/src`.
To build components package from the source code can be useful, for example, to enable tree shaking.

### DIFF
--- a/config/global.requires.js
+++ b/config/global.requires.js
@@ -1,7 +1,7 @@
 /**
  * Importing SCSS
  */
-import '@source/base/_index.scss'
+import '../src/base/_index.scss'
 
 /**
  * Importing Libraries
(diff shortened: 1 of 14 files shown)


sqevani:
I would say that importing from `src` is an extreme edge case, but either way I can see your point. 

For me just made more sense if you had a point of reference as a root folder and you went from there, instead of figuring out which level of the folder you were at. Also if you had to move folders around, it would be less troublesome refactoring every directory.

But lets go with your proposal.

PR #141, opened 2018-08-29
Drop unused mixins

davidyuk:


sqevani:
For the mixins, I would like to keep them there for now! there is no harm to them being there.

davidyuk:
You can add them when they become necessary. The harm that they make is that it becomes more difficult to understand what happens in the project.

sadiqevani:
Okay then, lets remove them.

PR #142, opened 2018-08-29
Fix styles folder

davidyuk:


sqevani:
I will keep this folder as it is for now.

davidyuk:
> I will keep this folder as it is for now.

`styles` is more accurate than `base`, why don't rename it?

> Also, the `_` prefix implies partials, its kinda of a standard in `scss`.

Yes, but we are using webpack and importing scss files explicitly, so these prefixes do not make any sense.

PR #143, opened 2018-08-29
Build to dist folder

davidyuk:
@sqevani why you have changed the folder for built files from `dist` to `build`? Most of our projects use `dist`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,6 +7,6 @@ package.json
 /index.js
 
 # Folders
-/build
+/dist
 /config
 /node_modules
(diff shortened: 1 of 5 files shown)


sqevani:
@davidyuk I think there is still a bit more changes that need to be done when it comes to build/folder.

Then I would say renaming the folders like this:

- `build` -> `dist` dist files
- `config` -> `build` build files

davidyuk:
But you have put the jest configs into `config` folder, that is not related to build process. After merging of #144 and #124 we won't have any stuff to put into `config`/`build` folder. Let's keep this PR only about the renaming of `build` to `dist`.

sadiqevani:
@davidyuk Okay then! So do you wanna proceed and merge?

PR #145, opened 2018-08-30
Make validators consistent

davidyuk:


davidyuk:
@sadiqevani Arrow functions should be enough obvious because they were introduced in ES6 four years ago. `object-shorthand` rule of Airbnb configs (that you have chosen in #160) rejects code like:
```js
const t = {
  validator: function() {
    // ...
  }
}
```
at least should be used:
```js
const t = {
  validator() {
    // ...
  }
}
```
but arrow functions, in this case, is better because it is shorter:
- no need to write `return`
- curly brackets are not necessary
- simple validators can be written in a single line

Also, functions starting with `() => ` without curly brackets are easier to read, because:
- it has only one statement
- it can't define internal variables
- it can't have own `this`

PR #149, opened 2018-09-03
AeDrop: Use @click instead of v-touch

davidyuk:


sqevani:
Any reasoning why you would like to change this back to click?

davidyuk:
As I see, `@click` does the same as `@tap` of VTouch component. Seems that the external package is not necessary in this case.

davidyuk:
@sadiqevani Ok, but if there any difference in behavior of `v-touch @tap` and `@click`? If not, so we shouldn't use v-touch in this case, because click/tap handling is already implemented in Vue.

davidyuk:
If `v-touch @tap` and `@click` have the same behavior it is better to use `@click` because in the case if a developer uses only AeDrop component and tree shaking is enabled then built bundle will be smaller because it won't contain `v-touch` and `hammerjs` packages.

davidyuk:
I can find only that `tap` fires faster than `click` when we don't have `"width=device-width"` viewport meta tag, but we have this tag in our apps.
https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away

PR #152, opened 2018-09-03
Add AeListItem from the Base app

davidyuk:
Based on https://github.com/aeternity/aepp-identity/blob/7151b2c/src/components/ListItem.vue

### DIFF
--- a/src/components/ae-list-item/ae-list-item.vue
+++ b/src/components/ae-list-item/ae-list-item.vue
@@ -1,57 +1,89 @@
 <template>
-    <li class="ae-list-item" :class="{ [fill]: Boolean(fill) }">
-      <slot />
-    </li>
+  <component
+    :is="renderAs"
+    :to="to"
+    :class="[fill, inactive && 'inactive']"
+    class="ae-list-item"
+    @click="$emit('click', $event)"
+  >
+    <!-- @slot List item content -->
+    <slot />
+    <div class="space" />
+    <!-- @slot List item content on the right side -->
+    <slot name="right" />
+  </component>
 </template>
+
 <script>
+import AeLink from '../aeLink/aeLink.vue';
+
 export default {
   name: 'ae-list-item',
   props: {
     /**
-     * Fill property changes the color of the borders.
-     * Select something between: `primary, secondary, neutral, alternative`
-     * fallback: `neutral`
+     * Path to go to when clicked
+     */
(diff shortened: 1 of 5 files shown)


src/base/_typography.scss
```diff
@@ -13,17 +13,3 @@ p {
   line-height: $base-line-height;
   font-size: $base-font-size;
 }
-
-a {
-  color: inherit;
-  transition: all $base-transition-time;
-  opacity: 0.75;
-  font-weight: bold;
```
davidyuk: These styles are too specific, I don't want to undo them in components like AeListItem one.

sqevani:
this does not make sense to me overall with the way I am thinking that components will interact with each other.

davidyuk:
> this does not make sense to me overall with the way I am thinking that components will interact with each other
So, please explain how components should interact with each other.

> The component should be stateless and with not change behavior on request
AeListItem is stateless (it doesn't have `data` defined). On what request AeListItem changes its behavior?

> later if we see things are not good, we can change them
This component is already used in the Base app,  not necessary to reinvent it.

src/components/ae-list-item/ae-list-item.vue
```diff
@@ -1,57 +1,89 @@
...
+  <component
+    :is="renderAs"
+    :to="to"
+    :class="[fill, inactive && 'inactive']"
+    class="ae-list-item"
+    @click="$emit('click', $event)"
+  >
+    <!-- @slot List item content -->
+    <slot />
+    <div class="space" />
```
sadiqevani, line 11: @davidyuk 

In here why are you inserting a `space` div, what is its specific functionality? and why is it needed?

davidyuk: In general, ListItem should be able to render elements on the left and the right sides. The same can be implemented using `justify-content: space-between` but then slots should be wrapped with divs because if we are passing several elements into the slot, they will be aligned along free space. `<div class="space" />` has a poor semantic, but I don't think that two wrapping divs are better.
Usage example: https://github.com/aeternity/aepp-identity/blob/be5f562a43e39545aafce46909db089c98c23ee5/src/pages/Settings.vue#L15-L21

sadiqevani:
Closing this based on the new version of this component: https://github.com/aeternity/aepp-components/pull/213

PR #153, opened 2018-09-03
Refactor AeCheck

davidyuk:
Expected to be used with AeListItem from #152 like in [Network settings screen](https://github.com/aeternity/aepp-identity/blob/c8232aedc92b011625ff28d5cac7f08d5eec262f/src/pages/SettingsNetwork.vue#L9-L19) of the Base app.
Restored `<input type="checkbox" />` behavior in Vue like in [this examples](https://vuejs.org/v2/guide/forms.html#Checkbox), according to [this guide](https://vuejs.org/v2/guide/components-custom-events.html#Customizing-Component-v-model).
Some changes based on [AeRadio](https://github.com/aeternity/aepp-identity/blob/5c787f1/src/components/AeRadio.vue) component from the Base app.

### DIFF
--- a/src/components/ae-check/ae-check.md
+++ b/src/components/ae-check/ae-check.md
@@ -7,26 +7,30 @@
 <template>
   <div>
     <h2>Radio</h2>
-    <label>
-      <ae-check v-model="singleChoice" value="1" type="radio" />
-      Radio button 1
-    </label>
-    <label>
-      <ae-check v-model="singleChoice" value="2" type="radio" />
-      Radio button 2
-    </label>
-    <label>value: {{ singleChoice }}</label>
+    <div>
+      <ae-check v-model="singleChoice" value="1" type="radio">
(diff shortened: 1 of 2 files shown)


sqevani:
Please don't modify styles that I've already done! If you have a proposal or change that makes sense would be best if there was a user-story first so I can find a solution based on what I was thinking about the components and its relationships with its overall parents/sibling components.

davidyuk:
> Components should be simple. Make them focus on a few simple things. Reuse other components inside your component. Design yours, so that other components can reuse it and create richer compositions.
https://github.com/aeternity/aepp-components#component-complexity

AeCheck should be as simple as `<input type="checkbox" />`, not need to combine this component with the functionality of AeListItem. Also the Base app already uses AeRadio components with a similar interface (together with AeListItem), I don't see any reason to change it.

PR #167, opened 2018-10-02
Add AePlainInput component

etharner:
Use this component instead of ae-text in contenteditable mode.

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -9,6 +9,7 @@ export { default as AeFlip } from './ae-flip/ae-flip.vue';
 export { default as AeIcon } from './ae-icon/ae-icon.vue';
 export { default as AeIdenticon } from './ae-identicon/ae-identicon.vue';
 export { default as AeInput } from './ae-input/ae-input.vue';
+export { default as AeInputPlain } from './ae-input-plain/ae-input-plain.vue';
 export { default as AeList } from './ae-list/ae-list.vue';
 export { default as AeListItem } from './ae-list-item/ae-list-item.vue';
 export { default as AeQRCode } from './ae-qrcode/ae-qrcode.vue';
--- a/src/styles/placeholders/typography.scss
+++ b/src/styles/placeholders/typography.scss
@@ -4,6 +4,9 @@
 // *****************************
 //------------------------------
 
+@import '../variables/typography.scss';
+@import '../globals/functions.scss';
+
 /// Typography Sans
 %face-sans-l {
   font-family: $font-sans;
--- a/src/components/ae-backdrop/ae-backdrop.md
+++ b/src/components/ae-backdrop/ae-backdrop.md
@@ -3,7 +3,7 @@
     <ae-card>
       <template slot="avatar">
         <ae-identicon address="ak$QY8VNEkhj7omMUjAvfVBq2NjTDy895LBYbk7qVxQo1qT8VqfE" />
-        <ae-text contenteditable>My Account</ae-text>
+        <ae-input-plain placeholder="Account name" value="My account" />
       </template>
       <template slot="header">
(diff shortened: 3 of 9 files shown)


sadiqevani:
Hey @etharner 

Can you give me an overview to what where the reasons you think having a `plain-input` component would be better than `contenteditable` property? 

In the standard design guide there are not plain inputs, and the only usage of the plain input is in the `card` component.

from what I know `contenteditable` behaves the same as if its an input when its enabled, and also provides less hassle and less components.

etharner:
Hey @sadiqevani
As I see:
  1) you use contenteditable not only in ae-card, but in ae-backdrop and ae-flip.
  2) it behave not quite as enabled input, because it allows to use many shortcuts, which can change text pretty much (for example, ctrl-b, ctrl-i for bold, italic, etc.). So you need to disable it somehow. It's overkill for the component that needs to be simple.                                
  3) you can't receive value from your ae-text component (it's not accessible with v-model and you are not implementing input event handling).

config/styleguidist.config.js
```diff
@@ -23,11 +23,27 @@ module.exports = {
    */
   title: require('../package.json').name + ' \n ' + 'Version: ' + require('../package.json').version,
 
-  /**
-   * Components folder
-   * TODO: later remove '-' for folder recognition
-   */
-  components: path.resolve(__dirname, '../src/components/**/*-*.vue'),
+  components: [
```
sadiqevani: This is a bit cumbersome to manually add components, can you revert back to the original code?

src/components/index.js
```diff
@@ -17,6 +17,7 @@ import AeListItem from './ae-list-item';
 import AeQRCode from './ae-qrcode';
 import AeText from './ae-text';
 import AeToolbar from './ae-toolbar';
+import AePlainInput from './aePlainInput';
```
sadiqevani: Rename the folder / files to: `ae-input-plain`

src/components/aePlainInput/aePlainInput.vue
```diff
@@ -0,0 +1,56 @@
...
+    placeholder: String,
+    value: undefined,
+  },
+};
+</script>
+
+<style lang="scss" scoped>
+@import '../../styles/placeholders/typography';
+@import '../../styles/variables/colors';
+@import '../../styles/variables/animations';
```
sadiqevani: Use this import for variables:

```
@import '../../styles/globals';
```

src/styles/placeholders/typography.scss
```diff
@@ -4,6 +4,8 @@
 // *****************************
 //------------------------------
 
+@import '../variables/typography.scss';
```
sadiqevani: There is no need to add this, all of the needed global variables are found in:

```
@import '../../styles/globals';
```

src/components/aePlainInput/index.js
```diff
@@ -0,0 +1,5 @@
+import AePlainInput from './aePlainInput.vue';
```
sadiqevani: The `index.js` file should look like this:

```
import AeBadge from './ae-badge.vue';

export const install = (Vue) => {
  Vue.component('ae-badge', AeBadge);
};

export default AeBadge;
```

etharner:
@sadiqevani I've changed stuff. Ready to merge?

sadiqevani:
@etharner 

Here is all the steps that you need to update your component for me to merge it:

- [ ] change the name to: `ae-input-plain`(this means everything that has that name-space needs to change, the css classes and so on)
- [ ] add a new line after: `@extend %face-sans-base;`
- [ ] Revert back to the changes of `ae-text` in case you want to add or remove something, make a new pull-requests for `ae-text`

PR #168, opened 2018-10-04
Drop `index.js` in every component folder

davidyuk:
Based on #124

### DIFF
--- a/config/component-wrapper-template.js
+++ b/config/component-wrapper-template.js
@@ -1,0 +1,7 @@
+import Component from 'COMPONENT_PATH';
+
+export const install = function (Vue) {
+  Vue.component(COMPONENT_NAME, Component);
+};
+
+export default Component;
--- a/config/webpack.components.js
+++ b/config/webpack.components.js
@@ -30,13 +30,28 @@ const components = fs.readdirSync(
  * Exporting components
  */
 module.exports = components.map((name) => merge(require('./webpack.config'), {
-  entry: path.resolve(__dirname, '../src/components', name, 'index.js'),
+  entry: path.resolve(__dirname, './component-wrapper-template.js'),
   output: {
     path: path.resolve(__dirname, '../dist', name),
     filename: 'index.js',
     libraryTarget: 'umd',
     library: '@aeternity/aepp-components/' + name,
     umdNamedDefine: true
+  },
+  module: {
+    rules: [{
+      include: path.resolve(__dirname, './component-wrapper-template.js'),
+      loader: 'string-replace-loader',
+      options: {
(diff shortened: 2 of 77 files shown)


sadiqevani:
Hey @davidyuk 

- Why have you removed the `index.js` I believed we agreed not to make major changes on the components.
- Naming new components with `[component name]2` is not a good practice, I would prefer to have the new component as the `[component name]` and the old component `[component name]Old` or something similar that makes sense, I leave that up to you
- for now lets not refactor things! I think the state of the project here is okay. What is needed mostly now are adding components rather than changing the structure.

davidyuk:
> Why have you removed the index.js

You mean in every component folder? They are not necessary, nothing is broken without them.

> Naming new components with [component name]2 is not a good practice

I agree that changes were too hasty, let's keep only yours components exported by default, but in the future, we need to merge the functionality of components with the same name.

> for now lets not refactor things! I think the state of the project here is okay.

This code is already implemented, it removes 450loc of useless code and prevents from a writing of useless code in future, you can't reject this PR just because it is refactoring.

sadiqevani:
Hey @davidyuk 

I'm not saying i'm rejecting it, I just don't agree with **some** of the changes.

- You mean in every component folder? They are not necessary, nothing is broken without them.

Nothing is broken, I know, but also functionality that I want to have in the component repository is being removed. The per-component export that I think solves some specific use-cases where developers can include only a single component are not being implemented.

Not everyone has implemented tree-shaking, and this is an open source project, so we need to support a wide use-case.

So the `index.js` acts as a middle man for different levels of export.

-----

I would like to postpone the merging of this pull-request for later, since I don't see any immediate value for now.

davidyuk:
> Nothing is broken

So, actually, I have broke per-component build, I forgot to check it. It is fixed in the latest commit.

> and this is an open source project, so we need to support a wide use-case

Most of the component packages are open source, have you seen/used a similar per-component build process anywhere else? Vue CLI is intentioned to be used for the building of components libraries, they propose to [build Vue components into async web components](https://cli.vuejs.org/guide/build-targets.html#async-web-component), I think it suits for the wider set of use-cases than your approach, and it can be configured much simpler, and it will be maintained by `vue-cli` team.

> I don't see any immediate value for now

Future PRs (for example #134) can be simpler because won't be necessary to write index.js file for every component. Postponing of this PR will demand resources to keep it up to date, to be able to merge it in future.

sadiqevani:
@davidyuk Lets merge this pull-request as well! I like the solution for the index.js per-component output.

davidyuk:
Sure! The branch is rebased on the top of `develop`.

PR #169, opened 2018-10-04
fixing jest-cli issue and including it as a local package

sadiqevani:
There is an issue with tests for people who don't have `jest-cli` installed globally, so i'm including it as a dev dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -280,6 +280,12 @@
         }
       }
     },
+    "acorn-walk": {
+      "version": "6.1.0",
+      "resolved": "https://registry.npmjs.org/acorn-walk/-/acorn-walk-6.1.0.tgz",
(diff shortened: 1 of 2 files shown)


davidyuk:
What particular issue do they have? I don't have jest installed globally, but `npm run unit` in components repository works well.

sadiqevani:
`jest-cli` wasn't being resolved in the `node_modules` folder.

davidyuk:
Is `npx jest` works without installing of `jest-cli`?

sadiqevani:
@davidyuk yep works.

I'll update the npm scripts to run jest through npx

sadiqevani:
@davidyuk seems the issue was within my `node_modules` folder, I'll close this pull-request.

PR #170, opened 2018-10-05
Drop AeIdentityAvatar, fix size bug in AeAvatar

davidyuk:


davidyuk:
@sadiqevani ready to merge?

PR #177, opened 2018-10-08
removing unnecessary items from backdrop

sadiqevani:
So, i've simplified the backdrop to remove align props, maybe we can add it later, but for now I think might be unnecessary.

But the component needs some improvement in the future, stylistic ones. Ray has designed this component with 4 layers to give it a specific effect.

screenshot 2018-10-08 16 14 00

There are 4 layers of the backdrop, some of the effects for now need some kind of polyfill, especially the grayscale of the elements behind the backdrop component, which can be achieved through `backdrop-filter` but seems is not supported in every browser.


src/components/ae-backdrop/ae-backdrop.md
```diff
@@ -62,8 +62,8 @@
         </ae-toolbar>
       </ae-card>
       <br />  
-      <ae-button @click="showBackdrop" face="round" extend>Show Backdrop</ae-button>
-      <ae-backdrop align="top" v-if="backDropVisible" @click.native.capture="showBackdrop">
+      <ae-button @click="showBackdrop" face="primary" extend>Show Backdrop</ae-button>
```
davidyuk: AeButton doesn't support `face` property equal to `primary`, can you rebase it on the top of develop branch avoiding this change?

sadiqevani: @davidyuk done.

davidyuk: you are almost there

PR #178, opened 2018-10-08
adding ae-slide as a component

sadiqevani:


src/components/ae-slide/ae-slide.vue
```diff
@@ -0,0 +1,114 @@
...
+      return Object.assign({
+        arrows: false,
+        infinite: false,
+      }, this.options);
+    },
+  },
+};
+</script>
+<style lang="scss" scoped>
+  @import '../../styles/globals';
```
davidyuk: please remove extra 2 space indents in `<styles>` block in consistency with `<script>`

src/components/ae-slide/ae-slide.vue
```diff
@@ -0,0 +1,114 @@
...
+     * @link https://github.com/lukaszflorczak/vue-agile#options
+     */
+    options: Object,
+
+    /**
+     * Creates a full screen slider, where the element tries to go
+     * 100% height and 100% width of the parent element, and places
+     * the round dots inside of the slider
+     */
+    fullscreen: Boolean,
```
davidyuk: what is the use case for this feature?

src/components/ae-slide/ae-slide.vue
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  .ae-slide.fullscreen /deep/ .agile__dots {
+    position: absolute;
+    right: 0;
+    left: 0;
+    bottom: 0;
+    height: auto;
+    padding: rem(20px) rem(32px);
+    z-index: $stack-5;
```
davidyuk: why exactly `$stack-5` instead of `1` or `5`? Actually `$stack-5` is a bad name, I can't understand when I should use it.

src/components/ae-slide/ae-slide.vue
```diff
@@ -0,0 +1,114 @@
...
+    justify-content: center;
+    align-items: center;
+    height: 20px;
+    margin: 0;
+    padding: 0;
+  }
+
+  .ae-slide /deep/ .agile__dots > .agile__dot {
+    @include size(10px);
+
```
davidyuk: Actually,
```
width: 10px;
height: 10px;
```
takes the same amount of lines of code, I don't see any reason to extract it into the mixin.

sadiqevani: Updated.

src/components/ae-slide/ae-slide.vue
```diff
@@ -0,0 +1,114 @@
...
+    list-style: none;
+    border-radius: 50%;
+    background: $color-neutral;
+    margin: 0 rem(2px);
+    transition: all $base-transition-time;
+    cursor: pointer;
+
+    > button {
```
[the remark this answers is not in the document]
sadiqevani: `.class.based` scss styles I usually put them in their own containers and don't nest too much.
`element based` styles i usually do it inside the class.

I don't like nesting too much! I use it sparingly.

src/components/ae-slide/ae-slide.vue
sadiqevani: Any ideas for a better name for the z-index stacks?

src/components/ae-slide/ae-slide.md
```diff
@@ -0,0 +1,65 @@
...
+              </ae-button>
+              <ae-button face="toolbar">
+                <ae-icon name="share" />
+                Share
+              </ae-button>
+            </div>
+          </ae-toolbar>
+        </ae-card>
+      </div>
+      <div>
```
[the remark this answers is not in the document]
sadiqevani: I think any tag is fine. Its getting all `children` and applying styles / css classes, so it can be any type of container.

src/components/ae-slide/ae-slide.md
```diff
@@ -0,0 +1,65 @@
+```jsx
+  <div>
+    <ae-slide>
```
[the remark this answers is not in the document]
sadiqevani: updated.

src/components/ae-slide/ae-slide.md
```diff
@@ -0,0 +1,65 @@
...
+                <ae-icon name="copy" />
+                Copy
+              </ae-button>
+              <ae-button face="toolbar">
+                <ae-icon name="share" />
+                Share
+              </ae-button>
+            </div>
+          </ae-toolbar>
+        </ae-card>
```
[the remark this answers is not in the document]
sadiqevani: @davidyuk I think this needs a pull-request on its own! Can you make an issue for this?

src/components/ae-slide/ae-slide.vue
```diff
@@ -0,0 +1,114 @@
+<template>
+  <agile class="ae-slide" :class="{ fullscreen }" :options="defaults">
+    <slot />
+  </agile>
+</template>
+<script>
```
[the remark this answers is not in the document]
sadiqevani: Hmm, I think how it is for now is good.

src/components/ae-slide/ae-slide.vue
davidyuk: We should drop `stack.scss` and use just numbers if it is necessary at all.

sadiqevani: @davidyuk I'm okay with that, but long-term I think we need to find a standard way for z-index, as we don't want overlapping modals, popups over tooltips to be all around!

Can we merge this, and I'm going to make a pull-request to drop the `stack.scss`

src/components/ae-slide/ae-slide.md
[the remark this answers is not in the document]
davidyuk: It is usage example, it should be as simple as possible. Can I put AeCard into AeSlider without wrapping div? If not, so there is something wrong with AeSlider or AeCard.

src/components/ae-slide/ae-slide.vue
davidyuk: > For the tutorial

You mean for onboarding screens? We don't need to put all of them into a slider, vue router is used there, it also supports transitions between pages.

So, for me still unclear how it can be used.

src/components/ae-slide/ae-slide.vue
davidyuk: This kind of issues should be solved by order of nodes in markup instead of `z-index`, no need to write `z-index: 9999999999999999;`. For example, modals can be placed at the end of markup then they won't overlap with other elements unless they are not using z-index.
Please, check the sources of the Base app, it has minimal usage of `z-index`.

sadiqevani:
@davidyuk 

For now this will act more like an interface, rather than a long-term component, and yes we will implement our own version of the carousel, but now there is a time limit.

So for now lets merge this and later on we will update it with our own version (internal change, not interface change)

edwarddikgale:
Hey guys a story exists in AE components project to upgrade this with our own version written using vue-touch or we agree to use vue-awesome-swiper depending on its limitations. Denis can you pass this for now as we have a note to manage it...

davidyuk:
@etharner is currently using `vue-awesome-swiper` in the Base app, no need to make a temporary wrapper around `vue-agile` or `vue-awesome-swiper`

sadiqevani:
@davidyuk I still am of the opinion that we should go ahead and merge this, and later we can engineer our own slider, but for now I am / was in need of a slider functionality with the current design, I don't want that every time I need a slider, have to restyle vue-agile or vue-awesome-swiper

PR #184, opened 2018-10-10
Clear global styles

davidyuk:
Today I have added new components package to the Base app, and I ran into a problem with global styles of this package: they break stuff in the Base app, especially styles for `<a>`.

So, I propose to define as less as possible styles globally to avoid this problem, and make components package more compatible with other components packages.

### DIFF
--- a/src/styles/rules.scss
+++ b/src/styles/rules.scss
@@ -14,7 +14,6 @@
 html, body {
   font-size: $base-font-size;
   line-height: $base-line-height;
-  caret-color: $color-error;
 
   /* Better Font Rendering */
   -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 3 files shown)


src/styles/rules.scss
```diff
@@ -13,30 +13,21 @@
...
 * {
   outline: none;
   outline: 0;
-  box-sizing: inherit;
 
   &:hover,
   &:focus,
   &:active {
     outline: none;
     outline: 0;
```
davidyuk: Maybe it is better to implement `focus` behavior firstly?
http://www.outlinenone.com/

sadiqevani:
I'm closing this on favor of this solution : https://github.com/aeternity/aepp-components/pull/194

PR #196, opened 2018-10-19
Feature/add features to ae address

sadiqevani:


src/components/ae-address/ae-address.vue
```diff
@@ -86,5 +95,31 @@ export default {
   > li {
     list-style: none;
   }
+
+  &.short {
+    grid-template-columns: 1fr 1fr 1fr 1fr 1fr;
```
davidyuk:
```suggestion
    grid-template-columns: repeat(5, 1fr);
```

src/directives/copy-to-clipboard.js
```diff
@@ -0,0 +1,17 @@
+import copy from 'clipboard-copy';
+
+/**
+ * The following directive attaches an event listener
+ * to the element, and copies the value provided
+ * to the directive
+ * @return {Object}
+ */
+export default {
+  inserted: (el, binding) => el.addEventListener('click', () => copy(binding.value).then(() => {
```
davidyuk: Why don't use `await` here?
```suggestion
  inserted: (el, binding) => el.addEventListener('click', async () => {
    await copy(binding.value);
```

src/directives/copy-to-clipboard.js
```diff
@@ -0,0 +1,17 @@
+import copy from 'clipboard-copy';
+
+/**
+ * The following directive attaches an event listener
+ * to the element, and copies the value provided
+ * to the directive
```
davidyuk: You have missed `v-copied-to-clipboard` class description here.

src/components/ae-address/ae-address.vue
```diff
@@ -1,5 +1,12 @@
 <template>
-  <ul class="ae-address" :style="{ gridGap: gap }" v-remove-spaces-on-copy>
+  <ul
+    ref="address"
+    class="ae-address"
+    :class="[ length, { copied } ]"
```
davidyuk: `copied` is not defined anywhere.

davidyuk (commented):
according to which designs the “address copied” label is implemented?

src/directives/copy-to-clipboard.js
sadiqevani: @davidyuk doesn't work.

src/components/ae-address/ae-address.vue
sadiqevani: updated

src/directives/copy-to-clipboard.js
sadiqevani: updated

src/components/ae-address/ae-address.vue
sadiqevani: removed

sadiqevani:
> according to which designs the “address copied” label is implemented?

There was a discussion with Ray and Emil, and decided to implement this feature. As it's needed on token-migration.

sadiqevani:
@davidyuk 

Also another thing we need to do, is to have the copy functionality dynamic, so we can enable/disable it, any issues/suggestions on how to handle this with a directive?

davidyuk (approved):

davidyuk:
@sadiqevani As a workaround, I can propose to disable `copyToClipboard` directive if the passed value is empty.

src/directives/copy-to-clipboard.js
davidyuk: Why you have set `regenerator` filed to false in `.babelrc`?
4a3f2c8e81504975d8fda56a285edabdfe794994

sadiqevani: @davidyuk I don't remember the reason, though thanks for the update. I'll merge the branch now.

sadiqevani:
> As a workaround, I can propose to disable copyToClipboard directive if the passed value is empty.

But the value passed is directly taken from the address value, should we have a secondary value for copying to clip board?

PR #198, opened 2018-10-24
Enhance/ae input data change

sadiqevani:
- adding style semantics for color input / focus
- changing the event propagation to use event mixin
- adding v-model on :value property to change dynamic value
- adding `slot` component inside if we need custom inputs

### DIFF
--- a/src/components/ae-input-plain/ae-input-plain.vue
+++ b/src/components/ae-input-plain/ae-input-plain.vue
@@ -3,13 +3,16 @@
     :class="['ae-input-plain', fill]"
     :placeholder="placeholder"
     :value="value"
-    @input="$emit('input', $event.target.value)"
+    @input="propagate"
   />
 </template>
 
 <script>
+import { events } from '../../mixins';
+
 export default {
   name: 'ae-input-plain',
+  mixins: [events],
   props: {
     /**
      * Fill property changes the color of placeholder and value.
--- a/src/components/ae-input-plain/ae-input-plain.vue
+++ b/src/components/ae-input-plain/ae-input-plain.vue
@@ -3,7 +3,7 @@
     :class="['ae-input-plain', fill]"
     :placeholder="placeholder"
     :value="value"
-    @input="propagateValue"
+    @input="propagateEventValue"
   />
 </template>
 
--- a/src/components/ae-input-plain/ae-input-plain.vue
+++ b/src/components/ae-input-plain/ae-input-plain.vue
@@ -3,7 +3,7 @@
     :class="['ae-input-plain', fill]"
     :placeholder="placeholder"
     :value="value"
-    @input="propagate"
+    @input="propagateValue"
   />
 </template>
 
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -19,7 +19,7 @@
         :class="{ aemount }"
         @focus="focus = true"
         @blur="focus = false"
-        @input="propagateValue"
+        @input="propagateEventValue"
       />
       <slot />
     </div>
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -21,6 +21,13 @@
         @input="propagateEventValue"
         v-if="!$slots.default && !$scopedSlots.default"
       />
+      <!--
+        @slot adds the ability to add your own
+        custom input elements, if used in combination
+        with `scoped slots` (see https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots)
+        gives you the parent `context` as a property where
+        you can access methods / properties
+      -->
       <slot
         :context="this"
         v-else
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -161,12 +161,17 @@ export default {
   background: transparent;
   border: none;
   outline: none;
+
+  &:only-child {
+    flex: 1 0;
+  }
 }
 
 .ae-input.aemount {
   @extend %face-mono-xl;
 
   text-align: center;
+  line-height: 1;
   font-weight: 300;
 }
 </style>
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -105,18 +105,18 @@ export default {
   }
 
   &.focus {
-    border-left: 2px solid $color-magenta;
-    caret-color: $color-magenta;
+    border-left: 2px solid $color-focus;
+    caret-color: $color-focus;
   }
   &.focus .ae-input-label {
-    color: $color-magenta;
+    color: $color-focus;
   }
   &.error {
-    border-left: 2px solid $color-red;
-    caret-color: $color-red;
+    border-left: 2px solid $color-error;
+    caret-color: $color-error;
   }
   &.error .ae-input-label {
-    color: $color-red;
+    color: $color-error;
   }
 
   &.focus .ae-input-label:after,
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -10,16 +10,16 @@
       </div>
       <!-- Input tag -->
       <input
-        v-model="value"
         v-if="!$slots.default"
-        :type="type"
         :id="id"
+        :value="value"
+        :type="type"
         :placeholder="placeholder"
         class="ae-input"
         :class="{ aemount }"
         @focus="focus = true"
         @blur="focus = false"
-        @input="propagate"
+        @input="propagateValue"
       />
       <slot />
     </div>
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -48,7 +48,7 @@ export default {
     placeholder: String,
 
     /**
-     * Default input element value
+     * Actual input element value
      */
     value: [String, Number],
 
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -10,7 +10,6 @@
       </div>
       <!-- Input tag -->
       <input
-        v-if="!$slots.default"
         :id="id"
         :value="value"
         :type="type"
@@ -20,8 +19,12 @@
         @focus="focus = true"
         @blur="focus = false"
         @input="propagateEventValue"
+        v-if="!$slots.default && !$scopedSlots.default"
       />
-      <slot />
+      <slot
+        :context="this"
+        v-else
+      />
     </div>
     <!-- @slot footer slot, used for adding elements below the input -->
     <slot name="footer" />
--- a/src/components/ae-input/ae-input.vue
+++ b/src/components/ae-input/ae-input.vue
@@ -9,14 +9,19 @@
         <slot name="header" />
       </div>
       <!-- Input tag -->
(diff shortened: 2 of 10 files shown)


src/components/ae-input/ae-input.vue
```diff
@@ -95,18 +105,18 @@ export default {
...
   }
   &.error {
-    border-left: 2px solid $color-error;
-    caret-color: $color-error;
+    border-left: 2px solid $color-red;
+    caret-color: $color-red;
   }
   &.error .ae-input-label {
-    color: $color-error;
+    color: $color-red;
```
davidyuk: I think `$color-error` has better semantic than `$color-red` for "error" state of the input.

src/mixins/events/index.js
```diff
@@ -10,8 +10,20 @@ export default {
...
       return this.$emit(event.type, event);
     },
+
+    /**
+     * This method help propagate event values to
+     * parents components in vue.
+     * Its useful when you have input children components
+     * and want to use v-model in the parent component
+     * @param {Event} event
+     * @return {*|default.methods}
```
davidyuk: What means `default.methods`? Actually `this.$emit` always returns `this` or instance of the current component.

src/mixins/events/index.js
```diff
@@ -10,8 +10,20 @@ export default {
...
+     */
+    propagateValue(event) {
+      return this.$emit(event.type, event.target.value);
```
davidyuk:
```suggestion
      this.$emit(event.type, event.target.value);
```
I don't see reasons to return anything from here. Don't write unused code.

src/styles/variables/colors.scss
```diff
@@ -51,5 +51,8 @@ $color-shadow-alpha-15: rgba(0, 33, 87, 0.15);
 $color-shadow-alpha-10: rgba(0, 33, 87, 0.15);
 
 // General Colors
-$color-black: #000000 !default;
-$color-white: #ffffff !default;
+$color-black:   #000000 !default;
+$color-white:   #ffffff !default;
+$color-red:     $color-error !default;
```
davidyuk: It is clearer to define `$color-error` as `$color-red` and `$color-red` as `#FF0D6A`, because later, when we will decide to change `$color-error` to green or so, it won't break the value of `$color-red` variable.

src/styles/variables/colors.scss
```diff
@@ -51,5 +51,8 @@ $color-shadow-alpha-15: rgba(0, 33, 87, 0.15);
...
+$color-magenta: $color-focus !default;
+$color-cyan:    $color-focus-alt !default;
```
davidyuk: How `!default` is supposed to be used?

src/components/ae-input/ae-input.vue
```diff
@@ -42,6 +47,11 @@ export default {
      */
     placeholder: String,
 
+    /**
+     * Default input element value
```
davidyuk: Is it really a default value? I think it is "current" or "actual" value.

src/components/ae-input-plain/ae-input-plain.vue
```diff
@@ -3,13 +3,16 @@
     :class="['ae-input-plain', fill]"
     :placeholder="placeholder"
     :value="value"
-    @input="$emit('input', $event.target.value)"
+    @input="propagateValue"
```
davidyuk: I don't think that we should extract `$emit('input', $event.target.value)`, looks like that `$event` is defined specially for this case.

davidyuk: At least would be better to make `propagateValue` a directive instead of the method of `events` mixin.

src/mixins/events/index.js
```diff
@@ -10,8 +10,20 @@ export default {
...
+
+    /**
+     * This method help propagate event values to
+     * parents components in vue.
+     * Its useful when you have input children components
+     * and want to use v-model in the parent component
+     * @param {Event} event
+     * @return {*|default.methods}
+     */
+    propagateValue(event) {
```
davidyuk: It should be either `propagate` and `propagateValue` or `propagateEvent` and `propagateEventValue`.

src/components/ae-input/ae-input.vue
```diff
@@ -9,14 +9,19 @@
...
+        :value="value"
+        :type="type"
+        :placeholder="placeholder"
+        class="ae-input"
+        :class="{ aemount }"
+        @focus="focus = true"
+        @blur="focus = false"
+        @input="propagateValue"
+      />
+      <slot />
```
davidyuk: How this is supposed to be used? Maybe we should make additional components instead? Passing anything to default slot shouldn't break properties of this component (`value`, `placeholder`, `focus` and other fields becomes useless).

davidyuk (dismissed):

davidyuk:
> `enhance/ae-input-data-change`

We are using git flow, so please check [this](https://danielkummer.github.io/git-flow-cheatsheet/index.html) and name branches accordingly.

src/components/ae-input/ae-input.vue
sadiqevani: @davidyuk yep, you're right! I kinda went and used search replace for this, and didn't notice the class.

src/components/ae-input-plain/ae-input-plain.vue
sadiqevani: @davidyuk maybe we can do this later! Its just a workaround to make `v-model` work

src/components/ae-input/ae-input.vue
sadiqevani: @davidyuk 

In this case is a functionality that we want! Because providing that flexibility means I can add different type of inputs / content in there and just inherit the styles and some functionality from the `ae-input` component.

There are some edge cases that this supports. And it actually does not break anything, since the values just don't get passed to the `slot` component.

src/mixins/events/index.js
sadiqevani: updated

src/components/ae-input/ae-input.vue
sadiqevani: updated

src/styles/variables/colors.scss
sadiqevani: In case there is an overwritten value on your side where you use `$color-red` in a different context.

src/styles/variables/colors.scss
sadiqevani: updated

src/mixins/events/index.js
sadiqevani: updated

src/mixins/events/index.js
sadiqevani: updated

sadiqevani:
@davidyuk 

Any update on this?

src/components/ae-input/ae-input.vue
davidyuk: I don't agree if you want to have a wrapper around an arbitrary input so extract a separate component that will not have `value`, `placeholder`, `focus` properties. And use this component in AeInput and any other similar components.

sadiqevani: @davidyuk 

A workaround to accessing the `parent` context is to use `scoped-slots`, in this way we can access anything passed to the parent element that has to be used inside the child slot.

This slot is for flexibility, while working with token-migration, there were cases where we needed a slightly different `<input />` element, and in my opinion having multiple `<input />` components is not that good of an idea.

Using `<slot />` will only be useful in edge cases, so normally the user would simply want to use the styles we've provided but in these edge cases the `<slot />` gives a better solution

davidyuk: > and in my opinion having multiple <input /> components is not that good of an idea

Six months or so developers of components package decided to don't mix different types of inputs in a single component. I think still we should work according to it.

davidyuk (commented):
Internally in components package, I think we should build a set of input components that should look similar (they all should have a bright border at the left side and so on), for this purpose I propose to extract an component that will be used as wrapper for other input, but this component shoulnd't implement itself as input to make it some simplier and isolated. This wrapper also can be exposed to users to allow them to build own inputs usign it.

Some examples:
AeInput:
```
<ae-input-wrapper>
  <input />
</ae-input-wrapper>

```

AeTextarea:
```
<ae-input-wrapper>
  <textarea />
</ae-input-wrapper>

Also, it's okay for me to use `scoped-slots`.

**Comment by sadiqevani** [comment]
@davidyuk 
> Internally in components package, I think we should build a set of input components that should look similar (they all should have a bright border at the left side and so on), for this purpose I propose to extract an component that will be used as wrapper for other input, but this component shoulnd't implement itself as input to make it some simplier and isolated. This wrapper also can be exposed to users to allow them to build own inputs usign it.

This is similarly what I've been thinking as well, but to implement this it will create some breaking changes.

What I have in mind currently it would be to have the following structure:

- `ae-input` would be the wrapper for every input, and wont provide any functionality, except style changes / modifications

each subsequent input would be the actual `<input />` tag, an example list of inputs would be:
- ae-input-plain
- ae-input-amount
- ae-input-domain
- ae-input-address
- ae-input-textarea
- ae-input-creditcard
- ae-input-search

Like this we can also deal with other similar type of components, like the `select` html tag.

But the thing I don't wanna do is to start creating wrappers for everything, I think the `<ae-input>` should provide a basic `<input>` functionality, and then the extra inputs are just enhancing the input.

**Comment by edwarddikgale** [comment]
@sadiqevani how would this proposal make breaking changes?

<ae-input-wrapper>
  <input />
</ae-input-wrapper>

Also I like this idea your propose

ae-input-plain
ae-input-amount
ae-input-domain ...

the only thing I can see that impacts this is what happens if we decide to remove or add a style to what the ae-input's have in common. Then it is a possible maintenance issue. Plus side is that its brainless to use. I would love to put this proposal to the team and let them vote...

**Comment by sadiqevani** [comment]
@edwarddikgale 

> how would this proposal make breaking changes?

There are a few stuff I need to investigate and see if it makes sense. 

For breaking stuff:
- There's Child to Parent communication, and how to handle that
  - This affects for example the `focus` state where the component has a border applied on the sides.
- `ae-input-plain` is already a component, I'll need to change this to a more minimalistic version, which will break stuff in the base app

But I need to investigate if its doable and how we can do it without messing up.

File: src/components/ae-slider/ae-slider.md [hunk 10]
```diff
@@ -4,7 +4,7 @@
       <div>
         <ae-card fill="primary">
           <template slot="avatar">
-            <ae-avatar address="ak$QY8VNEkhj7omMUjAvfVBq2NjTDy895LBYbk7qVxQo1qT8VqfE" />
+            <ae-identicon address="ak$QY8VNEkhj7omMUjAvfVBq2NjTDy895LBYbk7qVxQo1qT8VqfE" />
```
ricricucit: old Address style?

ricricucit (changes requested):

ricricucit (approved):

davidyuk (dismissed):
I don't understand that breaking changes will happen if we will implement the wrapper component as I proposing instead of messing a lot of stuff in AeInput.

It is clearer, for example, to compose AeInputAddress with AeInputWrapper instead of AeInput (that is completed component with its own functionality).

> to start creating wrappers for everything

I meant just this particular case, you can make this wrapper not available for users of this package.

PR #204, opened 2018-11-05
increasing padding on round button

sadiqevani:


davidyuk:
What is the reason for this change? Can you link corresponding designs?

sadiqevani:
@davidyuk there were inconsistencies on designs related to the padding here, the buttons when they were one word and had to wrap on their text, it looked to small. Asked Ray and made the update.

davidyuk (approved):

PR #205, opened 2018-11-07
improving address chunking, and copy-to-clipboard style for visibility

sadiqevani:
- Now the chunking of the address happens in the format of: `ak_` `xx` `xxx` | `xx`$
- Improve copy-to-clipboard style on click, changing opacity for visibility and font-color to a darker one

### DIFF
--- a/src/components/ae-address/ae-address.vue
+++ b/src/components/ae-address/ae-address.vue
@@ -68,7 +68,7 @@ export default {
      * @return {String[]}
      */
     chunked() {
-      return this.value.match(/.{1,3}/g);
+      return this.value.match(/^(\w{2}_)|.{2}(?=.{47,53})|.{2,3}/g);
     },
   },
 };
@@ -105,10 +105,6 @@ export default {
     }
   }
 
-  &.v-copied-to-clipboard {
-    opacity: 0.4;
-  }
-
   &.v-copied-to-clipboard:before {
     @extend %face-mono-base;
 
@@ -117,7 +113,8 @@ export default {
     justify-content: center;
     align-items: center;
     font-weight: 500;
-    background: rgba($color-neutral-positive-1, 0.25);
+    color: $color-neutral-negative-3;
+    background: rgba($color-neutral-positive-1, 0.9);
     position: absolute;
     top: 0;
     right: 0;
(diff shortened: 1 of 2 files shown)


src/components/ae-address/ae-address.vue
```diff
@@ -68,7 +68,7 @@ export default {
      * @return {String[]}
      */
     chunked() {
-      return this.value.match(/.{1,3}/g);
+      return this.value.match(/^(\w{2}_)|.{2}(?=.{47,53})|.{2,3}/g);
```
davidyuk: Is it intentional to split `ak_QY8VNEkhj7omMUjAvfVBq2NjTDy895LBYbk7qVxQo1qT8VqfE1` as `ak_ QY 8VN Ekh j7o mMU jAv fVB q2N jTD y89 5LB Ybk 7qV xQo 1qT 8Vq fE1`?
Brakets around `\w{2}_` looks useless.

davidyuk:
Should we implement the same chunking in address input component?

src/components/ae-address/ae-address.vue
sadiqevani: @davidyuk 

Yes, generally we want to have:

- `ak_` as 3 characters at the beginning
- first two characters
- end of the address needs to display either 2 or 3 chars

This is how its generally in the designs.

sadiqevani:
@davidyuk 

> Should we implement the same chunking in address input component?

I think the address chunking in input component will have a different mechanism, since in the input we also have to deal with styling of the address and also whitespaces.

But when I'll work on that input, we maybe can abstract that regex functionality so it works on both?

src/components/ae-address/ae-address.vue
sadiqevani: @davidyuk I checked the tests, and there seem to be an issue:

We need to group into a pair of two the first two characters after `ak_` this is based on the designs which wont change I think.

davidyuk: I have not seen this screen:
https://app.zeplin.io/project/59e4cf99fc0bb2f99a89551b/screen/5bd1cec7e9f15622d2d99784
So looks like it is the intentioned behavior.

davidyuk:
In general address formatting is implemented in AeAddressInput component, I don't think that it is reasonable to re-implement it using regexps.

This kind of chunking can make annoying flickers while the user types an address.

davidyuk (approved):
It looks okay with my edits

Issue #202, opened 2018-10-31
Copy to clipboard address w/ a different UX


sadiqevani:
Task: Copy to clipboard address w/ a different UX

Description: When coping an address implement a mechanism that requires a secondary input or a short waittime until the address is copied.

Also display the full address on hover but with default computer UI/forms.

davidyuk:
This task is unclear to me. For what reason secondary input or waittime is needed?

> Also display the address on hover but with default computer UI/forms.

Add `title` attribute to the root node of address component?

sadiqevani:
@davidyuk 

The idea is that the user might not want to copy the address when clicking, maybe there needs to be a throttle that waits a bit until the address gets copied. 

If you have the latest chrome version, and you have a mac, you might have noticed that to quit chrome now you need to press CMD + Q and keep Q pressed. Some similar mechanism.

> Add title attribute to the root node of address component?

Yes this was the aim.

PR #207, opened 2018-11-09
adding new address layout

sadiqevani:
- adding a new layout for the address
  - the layout displays the address fully without breaks or grids in between
- adding `title` property to display full address on hover based on https://github.com/aeternity/aepp-components/issues/202#issuecomment-435790621


dadaphl:
LGTM

dadaphl (approved):
LGTM

PR #210, opened 2018-11-16
changing files to be single-file, structuring styles to have fallback…

sadiqevani:
This pull request deals with the issue of having two separate components repository. The new npm package `@aeternity/aepp-components-3` does not solve our issue and its slow in release and is a bit confusing.

I've turned the old components into a single `.vue` component so the `scss` in there gets bundled with the per-folder component, and also gets included in the `aepp.components.css` file so we don't need to use two different npm packages for the components.

@davidyuk I haven't tested on any application, and also we should have only one single npm package so we can do better transition.


src/components/aeAddressInput/aeAddressInput.vue
```diff
@@ -8,7 +8,6 @@
     v-remove-spaces-on-copy
   />
 </template>
-
```
davidyuk, line 11: It was intentional, I don't agree.

davidyuk (dismissed):
All hex colors should be uppercase? Really? Check Zeplin or as we wrote before.

davidyuk:
> The new npm package @aeternity/aepp-components-3 does not solve our issue

That issues with this package do we have? It works great in the Base app

src/components/aeAddressInput/aeAddressInput.vue
sadiqevani: What was intentional? The line break?

sadiqevani:
@davidyuk in what sense it works great? This is not a long-term solution, are we suppose to have:

- `@aeternity/aepp-components-3`
- `@aeternity/aepp-components-4`
- `@aeternity/aepp-components-5` 

For each major version?

I know it works, its just not a proper solution! We will go through with proper release versions and we'll export old components through the releases to keep backward compatibility.

People working with components are kinda confused at what they should import and foremost some people don't know where are the new components.

src/components/aeAddressInput/aeAddressInput.vue
davidyuk: @sadiqevani yes

davidyuk:
`@aeternity/aepp-components-3` is a temporary solution for the Base app, later we will be able to publish v3 at `@aeternity/aepp-components`

src/components/aeAddressInput/aeAddressInput.vue
ricricucit: I don't understand this review...is it about the linebreak or hex colors?

ricricucit (commented):
I've tried, but this is too big of a PR for me to review without knowing the project very well. I also don't understand why there would be the need of having several `@aeternity/aepp-components` packages...I'd need more details to review...sorry.

ricricucit (approved):
👍

PR #213, opened 2018-11-20
Update PR: "Add AeListItem from the Base app" 

sadiqevani:
The code in the pull-request https://github.com/aeternity/aepp-components/pull/152 for the component ae-list-item is too specific and too narrow for its general purpose.

I don't want to couple the new components tightly with other old components. So I've improved the code a bit to make the list item more of a general scope and not as strict as it was in the previous pull request.

### DIFF
--- a/src/components/ae-list-item/ae-list-item.md
+++ b/src/components/ae-list-item/ae-list-item.md
@@ -1,7 +1,5 @@
 ```vue
-<ae-list-item>
-  Content
-</ae-list-item>
+<ae-list-item>Content</ae-list-item>
 ```
 
 ```vue
--- a/src/components/ae-list-item/ae-list-item.md
+++ b/src/components/ae-list-item/ae-list-item.md
@@ -1,5 +1,22 @@
-```jsx
-  <ae-list-item>
-    hello world
-  </ae-list-item>
+```vue
+<ae-list-item>
+  Content
+</ae-list-item>
+```
+
+```vue
+<ul>
+  <ae-list-item fill="primary">Primary</ae-list-item>
+  <ae-list-item fill="secondary">Secondary</ae-list-item>
+  <ae-list-item fill="neutral">Neutral</ae-list-item>
+  <ae-list-item fill="alternative">Alternative</ae-list-item>
+</ul>
+```
+
+```vue
+<div>
+  <ae-list-item tag="div">Hello</ae-list-item>
+  <ae-list-item tag="label">Hello</ae-list-item>
+  <ae-list-item tag="span">Hello</ae-list-item>
+</div>
 ```
--- a/src/components/ae-list-item/ae-list-item.vue
+++ b/src/components/ae-list-item/ae-list-item.vue
@@ -1,27 +1,62 @@
 <template>
-    <li class="ae-list-item" :class="{ [fill]: Boolean(fill) }">
-      <slot />
-    </li>
+  <component
+    class="ae-list-item"
+    :is="tag"
+    :class="[fill, disabled]"
+    :to="to"
+    @click="propagate"
+  >
+    <!-- @slot List item content -->
+    <slot />
+  </component>
 </template>
 <script>
+import { events } from '../../mixins';
+
 export default {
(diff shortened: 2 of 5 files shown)


ricricucit (approved):
👍 looks like a good idea to have "flexibility" on html tags

davidyuk (commented):
This component is broken, `to` won't work without `tag=router-link`.

So, extraction of this component wasn't succeeded. I will continue to use the internal version of ListItem in aepp-base repository.

sadiqevani:
@davidyuk 

I tried the component and works good with the `to` and `tag` attributes. Now what I can think of to improve this component, would be to apply `router-link` when `to` attribute gets applied on the component, but this will increase complexity.

davidyuk:
> Now what I can think of to improve this component, would be to apply router-link when to attribute gets applied on the component

It was already implemented, you shouldn't have removed this stuff. `to` prop without `router-link` is useless.

@edwarddikgale What happening here?

sadiqevani:
@davidyuk 

- `router-link` is not an element which is part of the aepp-components
- I agree that now `to` is kinda useless unless there is no `router-link` but at the same time, that property needs to work if and only if you have `router-link`
- Providing a `tag` prop to render the right element is the way to go, rather than making conditions on what element to use
- AeLink is an old element and its going to be deprecated later.

davidyuk:
Why AeLink is going to be deprecated? Because of this component in CamelCase? AeLink was a solution for the same question about dependency on RouterLink what was raised by Philipp a long time ago.

sadiqevani:
> Why AeLink is going to be deprecated? Because of this component in CamelCase? AeLink was a solution for the same question about dependency on RouterLink what was raised by Phillip a long time ago.

It's a useless component. And I'm assuming that the problem was raised by Philipp when `router-link` was used internally in components.

The components are not going to use internal `router-link` (the new ones), if there will be need to replace the component tag with a router-link, then we'll use a different approach.

- Either apply `router-link` as a parent element
- Implement a mechanism that replaces the component tag with `router-link` as was done in `ae-list-item`

davidyuk:
Actually, I forgot about it, can be really better to use AeLink instead of RouterLink, great that you have noticed it!

> Providing a tag prop to render the right element is the way to go, rather than making conditions on what element to use

This property makes this component too flexible, in my version I have defined a set of options that are actually used.

Let's add AeComponent as a wrapper around Component and rewrite all projects using it.

davidyuk:
AeLink was implemented to add `to` property of AeButton component without explicit dependency on vue-router. It was very convenient to use it, I don't know why you break  it.

davidyuk:
> Implement a mechanism that replaces the component tag with router-link as was done in ae-list-item

That if RouterLink component of the different routing package for vue (have you seen them somewhere?) will accept route in a different property than `to`? This approach won't work.

davidyuk:
So, to make a button as a hyperlink I should always wrap it with my own RouterLink adding useless nodes into dom and breaking semantic, thank you Sadi!

davidyuk:
So, to make a button as a hyperlink I should always wrap it with my own RouterLink adding useless dom nodes and breaking semantic, thank you Sadi!

@emil-apeunit @edwarddikgale @dadaphl you are talking that this guy should make decisions regarding the components package?

sadiqevani:
@davidyuk 

1. **Leave your temper and irony outside when you're talking, they're not welcomed.**
2. `router-link` has the following prop, if you feel inclined in wrapping `router-link` on top of a button: https://router.vuejs.org/api/#tag
3. If you're facing an issue, just write up an issue ticket on the components repository, your behaviour is not constructive, and does not fix anything.
4. HTML5 Standard says that block level elements can go inside `<a></a>`
5. If you're the guy who fights over "semantics", then you should also agree that based on accessibility, a LINK should never look/feel like a BUTTON!
    1. https://marcysutton.com/links-vs-buttons-in-modern-web-applications/
    2. http://www.webaxe.org/proper-use-buttons-links/
    3. https://www.sitepoint.com/community/t/anchors-without-href/6690

PR #216, opened 2018-11-26
add :active style of ae-button as :hover effekt for desktop support

emilsomm:
- use `:active` state style, lowering opacity, on button for its `:hover` state to support desktop use
- this is related to style issues on the https://github.com/aeternity/aepp-token-migration and https://github.com/aeternity/aepp-token-migration-landing  project.

### DIFF
--- a/src/components/ae-button/ae-button.vue
+++ b/src/components/ae-button/ae-button.vue
@@ -96,6 +96,7 @@ export default {
   &:hover,
   &:focus {
     text-decoration: none;
+    opacity: 0.85;
   }
 
   &:active:not(:disabled) {


sadiqevani:
@emilsomm 

I'm not sure if the `:hover` state should have an opacity, the issue here is that when the button is also clicked and remains `:focused` it will apply the opacity, which in Mobile versions might keep the state like that until the user clicks somewhere else.

I think for the button state we should talk with @raymundjacobs and have two different states:
- `:focused` state should maybe apply opacity or inner shadow
- `:hover` state should maybe apply an increased box-shadow

emilsomm:
OK, we can discuss this with  @raymundjacobs <https://github.com/raymundjacobs> today !

sadiqevani:
I'm closing this PR, we can come back at this issue later.

PR #218, opened 2018-11-29
adding phraser component

sadiqevani:
- adding ae-phraser styles

### DIFF
--- a/src/components/ae-phraser/ae-phraser.vue
+++ b/src/components/ae-phraser/ae-phraser.vue
@@ -1,13 +1,9 @@
 <template>
   <div class="ae-phraser" :class="{ error }">
-    <span
-      class="ae-phraser-error"
-      :class="{ active: Boolean(error) }"
-      v-if="error"
-    >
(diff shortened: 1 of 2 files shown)


src/components/ae-phraser/ae-phraser.vue
```diff
@@ -1,17 +1,64 @@
...
+  }
+}
+
+.ae-phraser-container {
   display: flex;
-  flex-direction: column;
+  align-items: flex-start;
+  flex-wrap: wrap;
+
+  /deep/ > .ae-badge {
```
ricricucit, line 54: I'm not sure about the use of `/deep/`.
As Joel H. points out in [these comments](https://stackoverflow.com/questions/25609678/what-do-deep-and-shadow-mean-in-a-css-selector), Chrome has since deprecated the /deep/ combinator, and it gives a syntax error in IE.

ricricucit (approved):
👍I think the double type acceptance is weird, but that's ok. Discussed via chat.

Issue #164, opened 2018-09-25
ae-icon to unflexible


emilsomm:
I would suggest to change the neutral option of the icon to `color: inherit;` and not `color: #000000;`
This would truly be neutral, and we wouldn't have to overwrite it later. Which can be difficult. 

On the explorer the components also need updating

see screenshot below from my latest contribution to the explorer. In this situation i would use the neutral option as described above to give it a relatable color.  

<img width="281" alt="bildschirmfoto 2018-09-25 um 16 57 17" src="https://user-images.githubusercontent.com/20213342/46024193-89b2af00-c0e6-11e8-974a-a1269883b2c3.png">

davidyuk:
In the current version of the components package we are using [`fill`](https://github.com/aeternity/aepp-components/blob/d315ebbcad7f63e3a64a64194a3111ff6d0451d0/src/components/aeIcon/aeIcon.scss#L14) property to change icon color, it doesn't support `inherit` value. It can be implemented in a future version since AeIcon will be based on font file instead of inlined SVGs.

sadiqevani:
See #219

PR #219, opened 2018-12-03
Adding black color to ae-icon, adding inherit value to inherit color …

sadiqevani:
…from parent element unless specificed, removing size property as per Denis request, making name prop required

also based off: https://github.com/aeternity/aepp-components/issues/164


ricricucit (approved):
👍

PR #234, opened 2019-01-07
adding fraction style component

sadiqevani:
@raymundjacobs tasked me to create a component for the fraction element in the designs.

### DIFF
--- a/src/components/ae-fraction/ae-fraction.vue
+++ b/src/components/ae-fraction/ae-fraction.vue
@@ -51,10 +51,9 @@ export default {
 
 .ae-fraction {
   display: inline-block;
+  font-size: $base-font-size;
   font-family: $font-sans;
   font-weight: 500;
-  font-style: normal;
-  font-stretch: normal;
   line-height: 0;
   word-spacing: -4px;
   color: inherit;
--- a/src/components/ae-fraction/ae-fraction.vue
+++ b/src/components/ae-fraction/ae-fraction.vue
@@ -56,7 +56,6 @@ export default {
   font-weight: 500;
   line-height: 0;
   word-spacing: -0.25rem;
-  color: inherit;
 
   &.primary {
     color: $color-primary;
--- a/src/components/ae-fraction/ae-fraction.vue
+++ b/src/components/ae-fraction/ae-fraction.vue
@@ -54,7 +54,6 @@ export default {
   font-size: $base-font-size;
   font-family: $font-sans;
   font-weight: 500;
-  line-height: 0;
   word-spacing: -0.25rem;
 
   &.primary {
--- a/src/components/ae-fraction/ae-fraction.vue
+++ b/src/components/ae-fraction/ae-fraction.vue
@@ -55,7 +55,7 @@ export default {
   font-family: $font-sans;
   font-weight: 500;
   line-height: 0;
-  word-spacing: -4px;
+  word-spacing: -0.25rem;
   color: inherit;
 
   &.primary {
--- a/src/components/ae-fraction/ae-fraction.vue
+++ b/src/components/ae-fraction/ae-fraction.vue
@@ -1,0 +1,81 @@
+<template>
+  <span class="ae-fraction" :class="[ fill ]">
+    <sup>{{ numerator }}</sup>
+    &frasl;
+    <sub>{{ denominator }}</sub>
+  </span>
+</template>
+<script>
+export default {
+  name: 'ae-fraction',
+  props: {
+    /**
+     * Fill property changes the color state of the fraction component.
+     * Select something between: `primary, secondary, neutral, alternative, white, black`
+     */
+    fill: {
+      type: String,
+      validator: value => [
+        'primary',
+        'secondary',
+        'neutral',
+        'alternative',
+        'white',
+        'black',
+      ].includes(value),
+    },
+
+    /**
+     * The numerator property defines
+     * the upper number in the fraction.
+     */
+    numerator: {
+      type: [Number, String],
+      default: 0,
+    },
+
+    /**
+     * The numerator property defines
+     * the lower number in the fraction.
+     */
+    denominator: {
+      type: [Number, String],
+      default: 0,
+    },
+  },
+};
+</script>
+<style lang="scss" scoped>
+@import '../../styles/variables/colors';
+@import '../../styles/variables/typography';
+
+.ae-fraction {
+  display: inline-block;
+  font-family: $font-sans;
+  font-weight: 500;
+  font-style: normal;
+  font-stretch: normal;
+  line-height: 0;
+  word-spacing: -4px;
+  color: inherit;
+
+  &.primary {
+    color: $color-primary;
+  }
+  &.secondary {
+    color: $color-secondary;
+  }
+  &.neutral {
+    color: $color-neutral;
+  }
+  &.alternative {
+    color: $color-alternative;
+  }
+  &.white {
+    color: $color-white;
+  }
+  &.black {
+    color: $color-black;
+  }
+}
+</style>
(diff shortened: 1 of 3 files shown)


src/components/ae-fraction/ae-fraction.vue
```diff
@@ -0,0 +1,81 @@
...
+@import '../../styles/variables/colors';
+@import '../../styles/variables/typography';
+
+.ae-fraction {
+  display: inline-block;
+  font-family: $font-sans;
+  font-weight: 500;
+  font-style: normal;
+  font-stretch: normal;
+  line-height: 0;
```
davidyuk: Why `line-height` is `0`?

src/components/ae-fraction/ae-fraction.vue
```diff
@@ -0,0 +1,81 @@
...
+    },
+  },
+};
+</script>
+<style lang="scss" scoped>
+@import '../../styles/variables/colors';
+@import '../../styles/variables/typography';
+
+.ae-fraction {
+  display: inline-block;
```
davidyuk, line 53: `sup` and `sub` are inline elements, isn't it? Looks like This component hasn't any `block` content.

src/components/ae-fraction/ae-fraction.vue
```diff
@@ -0,0 +1,81 @@
...
+  font-family: $font-sans;
+  font-weight: 500;
+  font-style: normal;
+  font-stretch: normal;
+  line-height: 0;
+  word-spacing: -4px;
```
davidyuk: Shouldn't word spacing be relative to the font size?

src/components/ae-fraction/ae-fraction.vue
```diff
@@ -0,0 +1,81 @@
...
+<style lang="scss" scoped>
+@import '../../styles/variables/colors';
+@import '../../styles/variables/typography';
+
+.ae-fraction {
+  display: inline-block;
+  font-family: $font-sans;
+  font-weight: 500;
+  font-style: normal;
+  font-stretch: normal;
```
davidyuk: These two lines look useless because they are setting default values.

src/components/ae-fraction/ae-fraction.vue
```diff
@@ -0,0 +1,81 @@
...
+  line-height: 0;
+  word-spacing: -4px;
+  color: inherit;
```
davidyuk: It isn't `inherit` by default?

src/components/ae-fraction/ae-fraction.vue
```diff
@@ -0,0 +1,81 @@
...
+     */
+    fill: {
+      type: String,
+      validator: value => [
+        'primary',
+        'secondary',
+        'neutral',
+        'alternative',
+        'white',
+        'black',
```
davidyuk, line 24: Isn't better to use more general colors for easier adjustments in future? For example, Bootstrap uses `light` and `dark`: https://getbootstrap.com/docs/4.2/components/alerts/

src/components/ae-fraction/ae-fraction.vue
sadiqevani: Just to enforce line-hight to 0

src/components/ae-fraction/ae-fraction.vue
sadiqevani: yeah, so?

src/components/ae-fraction/ae-fraction.vue
sadiqevani: Auto is by default, not inherit.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: These are fine, we're using this ones overall throughout the components.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: Just enforcing rules, so they don't get overwritten easily.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: explicitly added front-size.

src/components/ae-fraction/ae-fraction.vue
davidyuk: is it necessary? can we remove it?

src/components/ae-fraction/ae-fraction.vue
davidyuk: let's remove this line

src/components/ae-fraction/ae-fraction.vue
davidyuk: I think you should set this value in `em` and remove explicit font-size.

src/components/ae-fraction/ae-fraction.vue
davidyuk: In with case, this difference has sense? Can't we avoid setting of `color` property?

src/components/ae-fraction/ae-fraction.vue
sadiqevani: I would like the behaviour of this element to be inline-block, and not simply inline.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: Yes, there are high chances that the element line-hight can change in different context.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: Fixed, not `em` but `rem`

src/components/ae-fraction/ae-fraction.vue
sadiqevani: I removed, though would have liked it for this component to inherit parent color, but this works as well.

src/components/ae-fraction/ae-fraction.vue
davidyuk: But I wouldn't like, what is the reasons for inline-block in this case?

src/components/ae-fraction/ae-fraction.vue
davidyuk: Does line-height can break `<sup>` and `<sub>`? If not, we don't have to worry about line-height.

src/components/ae-fraction/ae-fraction.vue
davidyuk: Okay, but if somebody changes font-size of this element he has to adjust `word-spacing` because now it is relative to the global font size. If word-spacing be in em, then it will correspond to current font-size of this element, so probably no need to adjust word-spacing. It looks more reasonable, why don't use `em` in this case?

src/components/ae-fraction/ae-fraction.vue
sadiqevani: Its not a matter of like or not! The way I want this component to behave is as an `inline-block` and not an inline element. I would like it to have both horizontal and vertical spacing.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: This element does not need to have a dynamic word spacing! It does not need to be dependent on a parent element. So in this case `em` does not make sense.

src/components/ae-fraction/ae-fraction.vue
sadiqevani: removed.

src/components/ae-fraction/ae-fraction.vue
davidyuk: Please provide an example how it can be used.

emilsomm (approved):
To me this looks good. But one concern i have, is that you rely on the default behaviour of the `<sub>`and `<sup>` elements. These could be overwritten by global styles unless they are referenced here as well. Even if we might not do this ourselves, someone else might pull in a complete reset of default behaviour, or alter their behaviour in an unscoped component of their own. This might only be my opinion :)

Issue #197, opened 2018-10-23
ae-address copy to clipboard needs to be disabled on condition


sadiqevani:
https://github.com/aeternity/aepp-components/pull/196#issuecomment-431850706

sadiqevani:
See https://github.com/aeternity/aepp-components/pull/239

PR #239, opened 2019-01-08
conditional copy to clipboard, unless is set to true, copyToClipboard…

sadiqevani:
… is optional

Fixes: https://github.com/aeternity/aepp-components/issues/197

### DIFF
--- a/src/components/ae-address/ae-address.vue
+++ b/src/components/ae-address/ae-address.vue
@@ -77,7 +77,7 @@ export default {
      * When set to `true` enables copy to clipboard
      * when the user clicks/taps the address
      */
-    copyOnClick: Boolean,
+    enableCopyToClipboard: Boolean,
   },
   computed: {
     /**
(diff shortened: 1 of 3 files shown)


src/components/ae-address/ae-address.vue
```diff
@@ -72,8 +72,23 @@ export default {
      * accepts any valid css value, ex: `10px, 1rem, 50%, 10px 10px`
      */
     gap: String,
+
+    /**
+     * When set to `true` enables copy to clipboard
+     * when the user clicks/taps the address
+     */
+    copyOnClick: Boolean,
```
davidyuk: The name of this property should be the same as the name of corresponding mixin because they are doing the same thing: enabling copy on click to clipboard.

Also, maybe this property should be enabled by default?

sadiqevani: I agree on the name, I'll update it. As for the default behaviour, I believe disabled would be the main behaviour, since you don't want to enable that feature everywhere, and also it needs some style update.

davidyuk (approved):

PR #240, opened 2019-01-08
add ae-input-range

sadiqevani:


src/components/ae-input-range/ae-input-range.vue
```diff
@@ -0,0 +1,144 @@
...
+ */
+@mixin range-track($background-color) {
+  height: rem(1.83px);
+  width: 100%;
+  background: $background-color;
+  box-shadow: none;
+  border: none;
+  border-radius: 0;
+  cursor: pointer;
+  transition: all $base-transition-time;
```
davidyuk: what particular properties are going to be transitioned?

src/components/ae-input-range/ae-input-range.vue
```diff
@@ -0,0 +1,144 @@
...
+ * Mixin to set the range track styles
+ */
+@mixin range-track($background-color) {
+  height: rem(1.83px);
+  width: 100%;
+  background: $background-color;
+  box-shadow: none;
+  border: none;
+  border-radius: 0;
+  cursor: pointer;
```
davidyuk: are these styles really necessary? we have a simpler version: https://github.com/aeternity/aepp-base/blob/ad418a623369031bf825325520dffc7c9e6a24a5/src/components/AeInputRange.vue#L50-L54

src/components/ae-input-range/ae-input-range.vue
```diff
@@ -0,0 +1,144 @@
...
+        'neutral',
+        'alternative',
+      ].includes(value),
+      default: 'primary',
+    },
+
+    /**
+     * Value property of the range input element
+     */
+    value: [String, Number],
```
davidyuk, line 37: It concerned to be a good practice set a default value or mark property as required, why you have removed it from our version? https://github.com/aeternity/aepp-base/blob/ad418a623369031bf825325520dffc7c9e6a24a5/src/components/AeInputRange.vue#L15

davidyuk (commented):
The next time would be better to copy an existing component and make adjustments to it in separate commits, to make comparison easier.

src/components/ae-input-range/ae-input-range.vue
sadiqevani:
```
value
The input control's value. When specified in the HTML, this is the initial value, and from then on it can be altered or retrieved at any time using JavaScript to access HTMLInputElement.value. The value attribute is always optional.
```

The value is optional.

src/components/ae-input-range/ae-input-range.vue
sadiqevani: Yeah, just making sure this works on all browsers as intended.

src/components/ae-input-range/ae-input-range.vue
sadiqevani: Removed.

src/components/ae-input-range/ae-input-range.vue
davidyuk: Have you made any tests in different browsers? Does it really make sense?

sadiqevani: I can't test on IE, but Webkit, Safari, Chrome, FF, works okay. in -webkit the progress tracker does not work though, its a default behaviour from -webkit from what I read online.

sadiqevani:
> The next time would be better to copy an existing component and make adjustments to it in separate commits, to make comparison easier.

Will do this next time! Though there are no major changes to the component, except adding of the standard colors, and adding some props to the mixin!

kenodressel:
@sadiqevani where are we with this?

sadiqevani:
@kenodressel currently is up to @davidyuk to review this, and then we merge!

sadiqevani:
@davidyuk  I've removed some of the unnecessary CSS, now it should be quite close with the one in the base-app!

Though I've set some hight props, because in firefox and chrome the height of the element behaves weirdly.

src/components/ae-input-range/ae-input-range.vue
```diff
@@ -0,0 +1,142 @@
...
+
+.ae-input-range {
+  -webkit-appearance: none;
+  width: 100%;
+  cursor: pointer;
+  margin: 1rem 0;
+  height: rem(2px);
+
+  &:focus {
+    outline: none;
```
davidyuk, line 95: Outline already removed below.

src/components/ae-input-range/ae-input-range.vue
```diff
@@ -0,0 +1,142 @@
...
+          background: map-get($colors, thumb-color);
+          box-shadow: 0 0 8px map-get($colors, shadow-color);
+          border-radius: 50%;
+          cursor: pointer;
+        }
+      }
+    }
+  }
+
+  &:focus, &:active { outline: none; }
```
davidyuk, line 130: Is it necessary to hide outline on `active`? Looks enough to hide outline on `focus`: http://jsfiddle.net/ke4Low1h/.

PR #241, opened 2019-01-08
adding default styles for ae-panel

sadiqevani:


ricricucit (approved):
👍

src/components/ae-panel/ae-panel.vue
```diff
@@ -0,0 +1,22 @@
...
+<script>
+export default {
+  name: 'ae-panel',
+};
+</script>
+<style lang="scss" scoped>
+@import '../../styles/variables/colors';
+
+.ae-panel {
+  position: relative;
```
davidyuk, line 15: `relative` positioning should be enabled in special cases, I don't see reasons for it here.

src/components/ae-panel/ae-panel.vue
```diff
@@ -0,0 +1,22 @@
...
+  background: $color-white;
+  border-radius: 8px;
+  box-shadow: 0 0 8px $color-shadow-alpha-10;
+  padding: 1rem;
+  min-height: 1px;
```
davidyuk, line 20: In which case is it necessary?

src/components/ae-panel/ae-panel.vue
sadiqevani: Its just a default behaviour I would like the component to have! It makes sense in case you want to position elements relative to the panel, which in the future might happen.

src/components/ae-panel/ae-panel.vue
sadiqevani: The component doesn't get displayed in case its empty

src/components/ae-panel/ae-panel.vue
davidyuk: This is the reason why I won't use this component.

PR #251, opened 2019-01-25
AeIcon: Align middle by default

davidyuk:
As I see in designs AeIcon is aligned by middle in more cases than by the baseline (default value).

Example:
```
<ae-icon name="copy" />Ag<ae-icon name="copy" />
```
Before: 
screenshot from 2019-01-25 08-27-08
After: 
screenshot from 2019-01-25 08-26-50


sadiqevani (approved):
LGTM

PR #252, opened 2019-01-25
AeIcon: Make it inline-flex only when it necessary

davidyuk:
I have checked styleguide, looks it doesn't break any other components/examples.

### DIFF
--- a/src/components/ae-icon/ae-icon.vue
+++ b/src/components/ae-icon/ae-icon.vue
@@ -69,9 +69,6 @@ export default {
 @import '../../styles/variables/colors';
 
 .ae-icon {
-  display: inline-flex;
-  justify-content: center;
-  align-items: center;
   vertical-align: middle;
 
   &.primary {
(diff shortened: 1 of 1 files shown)


sadiqevani:
@davidyuk 

Lets put this as an extra prop change (extra class that can be applied), rather than making it exclusively for `primary`.

davidyuk:
Did you mean `round` instead of `primary`? Are this styles used somewhere else except `round` state? I can't find its usages.

sadiqevani:
@davidyuk 

I was thinking more of something like:

```html
<ae-icon flex-center />
```

```javascript
props: {
  flexCenter: Boolean
}
```

```css
.ae-icon.flex-center {}
```

Adding those behavior modifications as class, so its not specific to one style.

davidyuk:
No need to add this property, parent component can modify icon styles if it is really necessary, I don't see any other cases except `round`.

PR #255, opened 2019-02-08
copyToClipboard: Copy actual value

### LINKED ISSUE #319: User gets the first address when copy the second one
Login with ledger nano S, has 2 accounts. When select the second one and click button to copy the address,  paste the address which still show the first one.

1. ledger nano s
2. chrome 71.0.3578.98
3. Macbook

davidyuk:
closes https://github.com/aeternity/aepp-base/issues/319

### DIFF
--- a/src/directives/copyToClipboard.js
+++ b/src/directives/copyToClipboard.js
@@ -5,25 +5,30 @@ import copy from 'clipboard-copy';
  * to the element, and copies the value provided to the directive
  * @param {String|Boolean} when falsy value copyToClipboard is disabled
  * @class v-copied-to-clipboard
- * @return {Object}
  */
-export default {
-  inserted: (el, binding) => el.addEventListener('click', async () => {
-    /**
-     * In case the value is falsy, do not proceed with
-     * the copyToClipboard functionality
-     */
-    if (!binding.value) return;
+export default (el, binding) => {
+  if (!('copyToClipboard' in el.dataset)) {
+    el.addEventListener('click', async () => {
+      const value = el.dataset.copyToClipboard;
 
-    /**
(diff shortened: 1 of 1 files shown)


davidyuk:
I think tests fail because of the too old version of jest:
https://github.com/jsdom/jsdom/issues/961
https://www.npmjs.com/package/jest/v/21.2.1
it can be fixed by #124

sadiqevani:
So @davidyuk, I went and replicated the issue in the ticket you referenced. Now, what I see as an issue is that the data in the directive does not update upon VNode update.

Using the internal `binding` would be a better approach than touching the DOM, in this case now that the directive returns a function, the function will execute on both: `bind, update` 

Wouldn't this work the same as your code?

```javascript
export default (el, binding) => el.addEventListener('click', async () => {
  /**
   * In case the value is falsy, do not proceed with
   * the copyToClipboard functionality
   */
  if (!binding.value) return;

  /**
   * Await for copy to be executed and proceed
   * with normal flow
   */
  await copy(binding.value);
  el.classList.add('v-copied-to-clipboard');
  setTimeout(
    () => el.classList.remove('v-copied-to-clipboard'),
    500,
  );
});
```

davidyuk:
> Wouldn't this work the same as your code?

No, it will add an event listener for every emitting of `bind`, `update` events, it should somehow remove previously added `click` listeners.

I think better to add event listener once, and then to update a variable with the current value that should be copied when it necessary. The problem is that directives don't have any own instance and they can only modify the corresponding element. To use `dataset` in this case is the approach recommended by [vue documentation](https://vuejs.org/v2/guide/custom-directive.html#Directive-Hook-Arguments).

